PR #19, opened 2024-12-04
AM-1011: add RV64A instruction handlers

lzhu-pub:


kathlenemagnus (approved):

kathlenemagnus:
@lzhu-pub you just need to run `make clangformat` and check in those changes. Otherwise this is good to go!

lzhu-pub:
how can we merge this?

PR #20, opened 2024-12-08
SV32 translation test with basic PageTableWalk

skhanx86:
This PR introduces a basic implementation of the SV32 page table walk, following a test-driven approach. A test was written to validate the functionality, which led to defining the `sv32PageTableWalk` method for address translation. The current implementation of `sv32PageTableWalk` checks the validity of PTEs but does not include comprehensive checks for other conditions. This sets up a foundation for further enhancement in the page table walk process, with future iterations expected to include more advanced validations.


core/PageTableWalker.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+      public:
+        PageTableWalker();
+        uint32_t getPFN(uint32_t, AtlasState*);
+        uint32_t sv32PageTableWalk(uint32_t, uint32_t, AtlasState*);
+    };
+
+} // namespace atlas
+
+#include "PageTableWalker.cpp"
```
kathlenemagnus, line 129: Header shouldn't include a source file like this.

core/PageTableWalker.cpp
```diff
@@ -0,0 +1,43 @@
...
+namespace atlas
+{
+    PageTableWalker::PageTableWalker() {}
+
+    uint32_t PageTableWalker::sv32PageTableWalk(uint32_t virtAddr, uint32_t satpRegVal,
+                                                AtlasState* state)
+    {
+        Va32 va(virtAddr);
+        // TODO: getBaseAddrFromSatpReg();
+        uint32_t PDEaddresss = satpRegVal + va.vpn1() * PTE_SIZE;
```
kathlenemagnus, line 16: All of these values can be made `const`.

core/PageTableWalker.cpp
```diff
@@ -0,0 +1,43 @@
+//
+// Created by skhan on 11/20/24.
+//
+
+#include "../include/AtlasTypes.hpp"
+
+namespace atlas
+{
+    PageTableWalker::PageTableWalker() {}
```
kathlenemagnus, line 9: This can be defined in the header.

core/PageTableWalker.cpp
```diff
@@ -0,0 +1,43 @@
+//
+// Created by skhan on 11/20/24.
+//
+
+#include "../include/AtlasTypes.hpp"
```
kathlenemagnus, line 5: Should be including `PageTableWalker.hpp` here.

core/PageTableWalker.cpp
```diff
@@ -0,0 +1,43 @@
...
+        uint32_t pteAddress = ptbaseAddress + va.vpn0() * PTE_SIZE;
+        uint32_t PhyMembaseAddress = getPFN(pteAddress, state);
+        uint32_t PhyMemFrameAddress = PhyMembaseAddress + va.offset() * PTE_SIZE;
+        return PhyMemFrameAddress;
+    }
+
+    // convert this to template method, make uint32_t as template
+    uint32_t PageTableWalker::getPFN(uint32_t entryValues, AtlasState* state)
+    {
+        uint64_t pteValueFromMem = state->readMemory<uint64_t>(entryValues);
```
kathlenemagnus, line 27: `const`

include/AtlasTypes.hpp
```diff
@@ -20,6 +20,8 @@
 #define PT_ENTRIES_SV32 1024 // Sv32 has 1,024 entries in a page table
 #define PT_ENTRIES_SV39 512  // Sv39 has 512 entries in a page table
 
+#define PTE_SIZE 4
```
kathlenemagnus, line 23: I don't think this needs to be here. We should be able to just check for RV32 or RV64 to get the PTE size.

test/core/translate/Translate_test.cpp
```diff
@@ -3,11 +3,19 @@
 #include "include/AtlasTypes.hpp"
 #include "core/PageTableEntry.hpp"
 #include "core/PageTable.hpp"
+#include "core/PageTableWalker.hpp"
 #include <bitset>
 #include "sparta/utils/SpartaTester.hpp"
 
 class AtlasTranslateTester
 {
+    bool writeToMemory(uint64_t pa, uint64_t val)
```
kathlenemagnus, line 12: You can use the `AtlasState` method to write to memory now. It's been modified to use the PA.

core/PageTableWalker.hpp
```diff
@@ -0,0 +1,129 @@
+//
+// Created by skhan on 11/20/24.
+//
+namespace atlas
+{
+
+    struct Va32Bits
```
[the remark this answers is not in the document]
skhanx86: By creating the struct and unions, I am abstracting the code which extracts the vpn(s), offset from the Virtual Address.
In the approach you mentioned, as far as I am understanding, it will need creation of methods which will extract vpn(s), offset but those will be called by passing the va everytime to those. Instead, my appraoch just needs a line of code for converting uint32_t virtualAddress to corresponding VA struct, and then all the code to extract VPN(s) and offset is abstracted.

kathlenemagnus (approved):

PR #22, opened 2024-12-10
Replace AtlasRegister with Sparta registers

colby-nyce:


core/inst_handlers/inst_helpers.hpp
```diff
@@ -57,3 +57,39 @@ inline int64_t mulh(int64_t a, int64_t b)
...
+        static uint64_t mask(const uint64_t old_val, const uint64_t new_val)
+        {
+            // The 'Mask' template parameter is a bit mask that specifies which bits are writable.
+            // We need to preserve the 'old_val' bits that are not writable, and replace the
+            // writable bits with 'new_val'.
+            return (old_val & ~Mask) | (new_val & Mask);
+        }
+    };
+
+    template <> struct RegisterBitMask<0>;
```
kathlenemagnus, line 74: Does this template specialization need to be defined?

kathlenemagnus (approved):

core/inst_handlers/inst_helpers.hpp
colby-nyce: I left this here as a sanity check that the generated code / macros are not using this with a mask of 0 which is invalid. If that bug creeps in, this will stop compiling since it's undefined (forward declared here).

PR #36, opened 2025-01-02
adding f and d extesion

lzhu-pub:
I am pretty sure there are logical mistakes about types. But eyeballing and scrutinizing can only carry me as far as it is now.

### DIFF
--- a/core/AtlasInst.hpp
+++ b/core/AtlasInst.hpp
@@ -54,7 +54,9 @@ namespace atlas
 
         uint32_t getOpcodeSize() const { return opcode_size_; }
 
-        sparta::Register* getRs1()
+        uint64_t getRM() { return opcode_info_->getSpecialField(mavis::OpcodeInfo::SpecialField::RM); }
+
+        sparta::Register* & getRs1()
         {
             sparta_assert(rs1_, "Operand RS1 is a nullptr! " << *this);
             return rs1_;
@@ -66,7 +68,13 @@ namespace atlas
             return rs2_;
         }
 
-        sparta::Register* getRd()
+        sparta::Register* & getRs3()
+        {
+            sparta_assert(rs3_, "Operand RS2 is a nullptr! " << *this);
+            return rs3_;
+        }
+
+        sparta::Register* & getRd()
         {
             sparta_assert(rd_, "Operand RD is a nullptr! " << *this);
             return rd_;
@@ -75,6 +83,8 @@ namespace atlas
         bool hasRs1() const { return rs1_ != nullptr; }
 
         bool hasRs2() const { return rs2_ != nullptr; }
+
+        bool hasRs3() const { return rs3_ != nullptr; }
 
         bool hasRd() const { return rd_ != nullptr; }
 
@@ -98,6 +108,7 @@ namespace atlas
         // Registers
         sparta::Register* rs1_;
         sparta::Register* rs2_;
+        sparta::Register* rs3_;
         sparta::Register* rd_;
 
         ActionGroup inst_action_group_;
--- a/core/AtlasInst.hpp
+++ b/core/AtlasInst.hpp
@@ -54,9 +54,12 @@ namespace atlas
 
         uint32_t getOpcodeSize() const { return opcode_size_; }
 
-        uint64_t getRM() { return opcode_info_->getSpecialField(mavis::OpcodeInfo::SpecialField::RM); }
+        uint64_t getRM()
+        {
+            return opcode_info_->getSpecialField(mavis::OpcodeInfo::SpecialField::RM);
+        }
 
-        sparta::Register* & getRs1()
+        sparta::Register*& getRs1()
         {
             sparta_assert(rs1_, "Operand RS1 is a nullptr! " << *this);
             return rs1_;
@@ -68,13 +71,13 @@ namespace atlas
             return rs2_;
         }
 
-        sparta::Register* & getRs3()
+        sparta::Register*& getRs3()
         {
             sparta_assert(rs3_, "Operand RS2 is a nullptr! " << *this);
             return rs3_;
         }
 
-        sparta::Register* & getRd()
+        sparta::Register*& getRd()
         {
             sparta_assert(rd_, "Operand RD is a nullptr! " << *this);
             return rd_;
--- a/core/inst_handlers/a/inst_helpers.hpp
+++ b/core/inst_handlers/a/inst_helpers.hpp
@@ -31,7 +31,11 @@ namespace atlas
         static_assert(sizeof(RV) >= sizeof(SIZE));
 
         using Op =
+<<<<<<< HEAD:core/inst_handlers/a/inst_helpers.hpp
             typename std::conditional<std::is_same_v<RV, RV64>,
+=======
+            typename std::conditional<std::is_same<RV, RV64>::value,
+>>>>>>> 97db081 (clang-format):core/inst_handlers/rv64/a/inst_helpers.hpp
                                       typename std::conditional<U, uint64_t, int64_t>::type,
                                       typename std::conditional<U, uint32_t, int32_t>::type>::type;
 
--- a/core/inst_handlers/a/inst_helpers.hpp
+++ b/core/inst_handlers/a/inst_helpers.hpp
@@ -31,11 +31,7 @@ namespace atlas
         static_assert(sizeof(RV) >= sizeof(SIZE));
 
         using Op =
-<<<<<<< HEAD:core/inst_handlers/a/inst_helpers.hpp
-            typename std::conditional<std::is_same_v<RV, RV64>,
-=======
             typename std::conditional<std::is_same<RV, RV64>::value,
->>>>>>> 97db081 (clang-format):core/inst_handlers/rv64/a/inst_helpers.hpp
                                       typename std::conditional<U, uint64_t, int64_t>::type,
                                       typename std::conditional<U, uint32_t, int32_t>::type>::type;
 
--- a/core/inst_handlers/d/RvdInsts.cpp
+++ b/core/inst_handlers/d/RvdInsts.cpp
@@ -1,6 +1,11 @@
 #include "core/inst_handlers/d/RvdInsts.hpp"
+#include "core/inst_handlers/f/inst_helpers.hpp"
 #include "include/ActionTags.hpp"
 #include "core/ActionGroup.hpp"
+extern "C"
+{
+#include "source/include/softfloat.h"
+}
 
 namespace atlas
 {
@@ -15,17 +20,9 @@ namespace atlas
                 atlas::Action::createAction<&RvdInsts::fld_64_compute_address_handler, RvdInsts>(
                     nullptr, "fld", ActionTags::COMPUTE_ADDR_TAG));
             inst_handlers.emplace(
-                "flw",
-                atlas::Action::createAction<&RvdInsts::flw_64_compute_address_handler, RvdInsts>(
-                    nullptr, "flw", ActionTags::COMPUTE_ADDR_TAG));
-            inst_handlers.emplace(
                 "fsd",
                 atlas::Action::createAction<&RvdInsts::fsd_64_compute_address_handler, RvdInsts>(
                     nullptr, "fsd", ActionTags::COMPUTE_ADDR_TAG));
-            inst_handlers.emplace(
-                "fsw",
-                atlas::Action::createAction<&RvdInsts::fsw_64_compute_address_handler, RvdInsts>(
-                    nullptr, "fsw", ActionTags::COMPUTE_ADDR_TAG));
         }
         else if constexpr (std::is_same_v<XLEN, RV32>)
         {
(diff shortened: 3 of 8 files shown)


core/AtlasInst.hpp
```diff
@@ -54,7 +54,12 @@ namespace atlas
 
         uint32_t getOpcodeSize() const { return opcode_size_; }
 
-        sparta::Register* getRs1()
+        uint64_t getRM()
```
knute-mips, line 57: Method can be made `const`

core/AtlasInst.hpp
```diff
@@ -54,7 +54,12 @@ namespace atlas
...
+        {
+            return opcode_info_->getSpecialField(mavis::OpcodeInfo::SpecialField::RM);
+        }
+
+        sparta::Register*& getRs1()
```
knute-mips, line 62: Curious is there's any advantage in returning a reference pointer...

core/inst_handlers/a/inst_helpers.hpp
```diff
@@ -31,7 +31,7 @@ namespace atlas
         static_assert(sizeof(RV) >= sizeof(SIZE));
 
         using Op =
-            typename std::conditional<std::is_same_v<RV, RV64>,
+            typename std::conditional<std::is_same<RV, RV64>::value,
```
knute-mips, line 34: Why the change?

core/inst_handlers/d/RvdInsts.cpp
```diff
@@ -162,381 +147,174 @@ namespace atlas
...
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        softfloat_roundingMode = getRM(inst);
+        const uint64_t rs1_val = inst->getRs1()->dmiRead<uint64_t>();
+        const uint64_t rs2_val = inst->getRs2()->dmiRead<uint64_t>();
+        const uint64_t rs3_val = inst->getRs3()->dmiRead<uint64_t>();
+        const uint64_t product = f64_mul(float64_t{rs1_val}, float64_t{rs2_val}).v;
+        inst->getRd()->write(f64_add(float64_t{product ^ 1UL << 63}, float64_t{rs3_val}).v);
```
knute-mips, line 190: Suggest `()` around the binary operations.

knute-mips: Curious... softfloat doesn't have an `madd` operations?  The way this is implemented might lose some precision.

core/inst_handlers/d/RvdInsts.cpp
```diff
@@ -162,381 +147,174 @@ namespace atlas
...
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        softfloat_roundingMode = getRM(inst);
+        const uint64_t rs1_val = inst->getRs1()->dmiRead<uint64_t>();
+        const uint64_t rs2_val = inst->getRs2()->dmiRead<uint64_t>();
+        const uint64_t rs3_val = inst->getRs3()->dmiRead<uint64_t>();
+        const uint64_t product = f64_mul(float64_t{rs1_val}, float64_t{rs2_val}).v;
+        inst->getRd()->write(f64_sub(float64_t{product ^ 1UL << 63}, float64_t{rs3_val}).v);
```
knute-mips, line 242: Suggest `()` around the binary operations.

core/inst_handlers/f/RvfInsts.cpp
```diff
@@ -222,434 +196,236 @@ namespace atlas
...
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        softfloat_roundingMode = getRM(inst);
+        const uint32_t rs1_val = inst->getRs1()->dmiRead<uint64_t>();
+        const uint32_t rs2_val = inst->getRs2()->dmiRead<uint64_t>();
+        const uint32_t rs3_val = inst->getRs3()->dmiRead<uint64_t>();
+        const uint32_t product = f32_mul(float32_t{rs1_val}, float32_t{rs2_val}).v;
+        inst->getRd()->write(f32_sub(float32_t{product ^ 1 << 31}, float32_t{rs3_val}).v);
```
knute-mips, line 251: Suggested `()`

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "core/AtlasState.hpp"
+#include "core/AtlasInst.hpp"
+#include "include/AtlasUtils.hpp"
+#include "mavis/OpcodeInfo.h"
+
+namespace atlas
+{
+    inline uint_fast8_t getRM(const AtlasInstPtr & inst)
+    {
+        uint64_t static_rm = inst->getRM();
```
knute-mips, line 12: `const`

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,60 @@
...
+        if (static_rm == 7) // RM field "DYN"
+        {
+            // TODO: use fcsr::frm
+            return 0;
+        }
+        else
```
knute-mips, line 18: Remove `else` and just `return static_rm`

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "core/AtlasInst.hpp"
+#include "include/AtlasUtils.hpp"
+#include "mavis/OpcodeInfo.h"
+
+namespace atlas
+{
+    inline uint_fast8_t getRM(const AtlasInstPtr & inst)
+    {
+        uint64_t static_rm = inst->getRM();
+        if (static_rm == 7) // RM field "DYN"
```
knute-mips, line 13: Why 7?

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,60 @@
...
+        }
+        else
+        {
+            return static_rm;
+        }
+    }
+
+    template <typename T> ActionGroup* compute_address_handler(AtlasState* state)
+    {
+        static_assert(std::is_same<T, RV64>::value || std::is_same<T, RV32>::value);
```
knute-mips, line 26: `std::is_same_v`

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        const T rs1_val = inst->getRs1()->dmiRead<uint64_t>();
+        constexpr uint32_t IMM_SIZE = 12;
+        const T imm = inst->hasImmediate() ? inst->getSignExtendedImmediate<T, IMM_SIZE>() : 0;
+        const T vaddr = rs1_val + imm;
+        state->getTranslationState()->makeTranslationRequest(vaddr, sizeof(T));
+        return nullptr;
```
knute-mips, line 34: Is this function a simple placeholder?  If it always returns `nullptr`, then why return anything?

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    using S = uint32_t;
+    using D = uint64_t;
```
knute-mips, line 38: This makes me nervous to see in a header file and in the global atlas namespace.  Are these types just for convenience?

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+    template <typename RV, typename SIZE>
+    ActionGroup* float_ls_handler(atlas::AtlasState* state, bool load)
+    {
+        static_assert(std::is_same<RV, RV64>::value || std::is_same<RV, RV32>::value);
+        static_assert(std::is_same<SIZE, S>::value || std::is_same<SIZE, D>::value);
```
knute-mips, line 44: `std::is_same_v`

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,60 @@
...
+        const T vaddr = rs1_val + imm;
+        state->getTranslationState()->makeTranslationRequest(vaddr, sizeof(T));
+        return nullptr;
+    }
+
+    using S = uint32_t;
+    using D = uint64_t;
+
+    template <typename RV, typename SIZE>
+    ActionGroup* float_ls_handler(atlas::AtlasState* state, bool load)
```
knute-mips, line 41: Can the parameter `load` be a template parameter?

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,60 @@
...
+    {
+        static_assert(std::is_same<RV, RV64>::value || std::is_same<RV, RV32>::value);
+        static_assert(std::is_same<SIZE, S>::value || std::is_same<SIZE, D>::value);
+        static_assert(sizeof(RV) >= sizeof(SIZE));
+
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        const RV paddr = state->getTranslationState()->getTranslationResult().getPaddr();
+
+        if (load)
```
knute-mips, line 50: If `load` is a template parameters, this can be `if constexpr`

core/inst_handlers/f/RvfInsts.cpp
```diff
@@ -222,434 +196,236 @@ namespace atlas
...
-
-        // require_extension('F');
-        // require_fp;
-        // WRITE_FRD(f32(RS1));
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        const uint64_t rs1_val = inst->getRs1()->dmiRead<uint64_t>();
+        inst->getRd()->write(rs1_val & 0x0000FFFF);
```
kathlenemagnus-mips, line 223: Can we define this mask in `AtlasTypes.hpp`? I'd prefer to avoid using any hard-coded values.

core/inst_handlers/d/RvdInsts.cpp
```diff
@@ -162,381 +147,174 @@ namespace atlas
...
-        // require_either_extension('D', EXT_ZDINX);
-        // require_fp;
-        // softfloat_roundingMode = RM;
-        // WRITE_FRD_D(i32_to_f64((int32_t)RS1));
-        // set_fp_exceptions;
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        const uint32_t rs1_val = inst->getRs1()->dmiRead<uint64_t>();
```
kathlenemagnus-mips, line 151: We should use the register macros defined here: https://github.com/sparcians/atlas/blob/main/arch/register_macros.hpp

@colby-nyce how do you recommend we use the register read and write macros in the instruction handler? Currently they take a string (the register name), but we don't have a method in `AtlasInst` to get the operand names for integer and floating point registers. We could add a method to `AtlasInst` to get the register number. Something like `getRs1Number`.

core/inst_handlers/a/inst_helpers.hpp
lzhu-mips: I think `value` is more compliant with the rest of the code with `type`.

core/inst_handlers/f/inst_helpers.hpp
lzhu-mips: hmn, which const?  making `static_rm` const?

core/inst_handlers/f/inst_helpers.hpp
lzhu-mips: from spec, if RM field is 2b'111 (DYN), then it's using csr to decide rounding mode.

core/inst_handlers/f/inst_helpers.hpp
lzhu-mips: thanks for pointing out.

core/inst_handlers/f/inst_helpers.hpp
lzhu-mips: @kathlenemagnus might have better answer for this. But the work done here is `makeTranslationRequest`, then at somewhere later `getTranslationResult` would return the translated PA.

core/inst_handlers/f/inst_helpers.hpp
lzhu-mips: it's for clarity. S = `single precision`. D = `double precision`. so when using the template, it won't be confused with RV64 vs RV32, which are again, uint32_t and uint64_t.

core/inst_handlers/f/inst_helpers.hpp
knute-mips: Yep

core/inst_handlers/a/inst_helpers.hpp
knute-mips: Ah.  I suggest fixing the other code.  ;)

core/inst_handlers/d/RvdInsts.cpp
colby-nyce: IMO, I have only used / felt the macros to be useful when writing to CSRs so we can deal with the bitmasks automatically. But I don't see the point in using the macros for rs1/rs2/rd. What would the call site look like and how is it better than what we are using today with getRs1/dmiRead directly?

If you want to use the macros, I would probably want to add new macros like READ_RS1_VAL, READ_RD_VAL, WRITE_RD_VAL, etc. 

typedef READ_RS1_VAL state->getCurrentInst()->getRs1()->dmiRead<uint64_t>()
...

But that isn't giving any real added value like with CSRs and bitmasks.

colby-nyce (approved):

core/inst_handlers/d/RvdInsts.cpp
kathlenemagnus-mips: I thought we needed the macros to control treating x0 as read-only. Is that true? Or is that handled in the Sparta register class?

If we only want to use the macros for the CSRs, then we should remove the integer and floating point macros.

PR #38, opened 2025-01-08
CSR/Exception unit

colby-nyce:


core/observers/InstructionLogger.cpp
```diff
@@ -64,6 +64,18 @@ namespace atlas
         AtlasInstPtr inst = state->getCurrentInst();
         sparta_assert(inst != nullptr, "Instruction is not valid for logging!");
 
+        if (!inst->hasRd()) {
+            INSTLOG("core   0: " << (uint64_t)state->getPrivMode() /*<< " " << HEX16(state->getPc())*/
+                                 << " " << "(" << HEX8(inst->getOpcode()) << ")");
+        } else {
+            INSTLOG("core   0: " << (uint64_t)state->getPrivMode() /*<< " " << HEX16(state->getPc())*/
```
kathlenemagnus-mips, line 71: I feel like this was probably added for CSR debug. Can it be removed since we already print the RD register values and previous values on line 110? We can keep the privilege mode print, but it doesn't need an if-statement.

core/inst_handlers/inst_helpers.hpp
```diff
@@ -92,4 +92,10 @@ namespace atlas
         }
     };
 
+#define get_field(reg, mask) \
```
kathlenemagnus-mips, line 95: How is this different from `GET_CSR_FIELD` and `WRITE_CSR_FIELD`? Do we need both?

core/AtlasState.cpp
```diff
@@ -128,4 +137,11 @@ namespace atlas
 
         return nullptr;
     }
+
+    uint64_t AtlasState::getMStatusInitialValue(const AtlasState* state, const uint64_t xlen_val)
```
kathlenemagnus-mips, line 141: This is fine for now, but we should do this type of initialization at the simulation level with a command line option like this:
```
--reg-init-val mstatus 42949672960
```

core/AtlasState.cpp
```diff
@@ -115,6 +121,9 @@ namespace atlas
 
         // Increment PC and upate simulation state
         action_group->insertActionAfter(increment_pc_action_, ActionTags::EXECUTE_TAG);
+
+        // Allow the Exception unit to deal with any unhandled exceptions
+        exception_unit_->insertExecuteActions(action_group);
```
kathlenemagnus-mips, line 126: For 99% of instructions, this Action will be useless so it would be better to redirect the instruction handlers to the Exception unit's Action only when an exception occurs.

So in your `TRAP_IMPL` macro, instead of returning a nullptr, you would do this:
```
#define TRAP_IMPL(cause)                                     \
    {                                                        \
        auto exception_unit = state->getExceptionUnit();     \
        exception_unit->setUnhandledException(cause);        \
        return exception_unit->getActionGroup();             \
    }
```

core/AtlasState.cpp
```diff
@@ -48,6 +49,11 @@ namespace atlas
         fetch_unit_ = getContainer()->getChild("fetch")->getResourceAs<Fetch*>();
         execute_unit_ = getContainer()->getChild("execute")->getResourceAs<Execute*>();
         translate_unit_ = getContainer()->getChild("translate")->getResourceAs<Translate*>();
+        if (auto exc = getContainer()->getChild("exception", false)) {
```
kathlenemagnus-mips, line 52: I'm assuming you added this because the regression tests were failing. Please add a FIXME comment for us to update the tests later.

core/Exception.cpp
```diff
@@ -0,0 +1,114 @@
...
+    (void)state;
+}
+
+void Exception::handleSModeException_(atlas::AtlasState* state)
+{
+    // TODO
+    (void)state;
+}
+
+void Exception::handleMModeException_(atlas::AtlasState* state)
```
kathlenemagnus-mips, line 60: This looks great! I think you'll find that these methods can be combined into a templated method, but this is good for now. Just add asserts to `handleSModeException_` and `handleUModeException_` to make sure Atlas fails if they get called.

core/inst_handlers/rv64/zicsr/RvzicsrInsts.cpp
```diff
@@ -28,116 +34,151 @@ namespace atlas
...
-        // }
-        // WRITE_RD(sext_xlen(old));
-        // serialize();
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & insn = state->getCurrentInst();
+
+        const auto & rs1 = insn->getRs1();
+        const auto rs1_val = rs1->dmiRead<uint64_t>();
```
lzhu-pub, line 40: I only use dimRead for memory accesses. Need some guideline here.

core/Exception.cpp
```diff
@@ -0,0 +1,114 @@
...
+    const reg_t trap_handler_address = (READ_CSR_REG(MTVEC) & ~(reg_t)1);
+    state->setNextPc(trap_handler_address);
+
+    const reg_t epc = state->getPc();
+    WRITE_CSR_REG(MEPC, epc);
+
+    const uint64_t cause = static_cast<uint64_t>(cause_.getValue());
+    WRITE_CSR_REG(MCAUSE, cause);
+    const uint64_t read_cause = READ_CSR_REG(MCAUSE);
+    (void)read_cause;
```
lzhu-pub, line 71: A re these for debugging conveniences?

core/inst_handlers/inst_helpers.hpp
colby-nyce: I'll remove both of these - they were debug utilities for comparisons against Spike that I forgot to revert.

core/AtlasState.cpp
colby-nyce: As we discussed, I'll leave this larger task as a follow-up after checking in all the Atlas IDE code first (especially the backend).

core/Exception.cpp
colby-nyce: Ah yes, the last two lines were leftover from debugging. I'll remove them.

core/AtlasState.cpp
colby-nyce: Done

core/inst_handlers/rv64/zicsr/RvzicsrInsts.cpp
colby-nyce: I know we mentioned adding some macro like READ_RS1 but other than making the code easier to read, it doesn't have a lot of benefits. I can change all the code to use macros later.

kathlenemagnus (approved):

PR #39, opened 2025-01-13
CSRs, Exception Unit, and Atlas IDE

colby-nyce:
I decided to check in everything at once including the Atlas IDE support. I closed the other PR for CSRs/Exception Unit since that's duplicated here.

### DIFF
--- a/core/AtlasState.cpp
+++ b/core/AtlasState.cpp
@@ -3,6 +3,7 @@
 #include "core/Fetch.hpp"
 #include "core/Execute.hpp"
 #include "core/Translate.hpp"
+#include "core/Exception.hpp"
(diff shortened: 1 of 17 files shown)


core/AtlasState.cpp
```diff
@@ -128,4 +153,199 @@ namespace atlas
...
+        uint64_t rd_val_before_;
+        uint64_t rd_val_after_;
+        int has_imm_;
+        uint64_t imm_;
+        std::string disasm_;
+        std::string mnemonic_;
+        uint64_t opcode_;
+        uint64_t pc_;
+        int priv_;
+        int result_code_;
```
knute-mips, line 279: Can these be `const`?

PR #40, opened 2025-01-13
CSRs, exception unit, and Atlas IDE backend/frontend

colby-nyce:


core/observers/CoSimObserver.hpp
```diff
@@ -20,6 +20,12 @@ namespace atlas
             return last_event_;
         }
 
+        void stopSim() override
```
kathlenemagnus, line 23: I'm not sure why this is needed since these values are set in the post-execute Observer method:
https://github.com/sparcians/atlas/blob/af0ad3343dd3258b3759743cb50fd78ec0612501/core/observers/CoSimObserver.cpp#L90-L91

Have you introduced a new way of stopping simulation?

core/AtlasState.cpp
```diff
@@ -128,4 +267,273 @@ namespace atlas
...
+                case 3:
+                    getCsrRegisterSet()->getRegister(reg_id)->dmiWrite(expected_val);
+                    break;
+                default:
+                    throw sparta::SpartaException("Invalid group num!");
+            }
+        }
+    }
+
+    class InstSnapshotter : public simdb::WorkerTask
```
kathlenemagnus, line 324: I'd prefer this class def to be in it's own file.

core/AtlasState.cpp
```diff
@@ -128,4 +267,273 @@ namespace atlas
...
         return nullptr;
     }
+
+    void AtlasState::enableCoSimDebugger(std::shared_ptr<simdb::ObjectManager> db,
+                                         std::shared_ptr<CoSimQuery> query,
+                                         const std::vector<RegisterInfo> &reg_info)
+    {
+        bool valid = inst_logger_.enabled();
+        if (!valid) {
+            for (auto obs : observers_) {
```
kathlenemagnus, line 277: Why do we need to enable all observers?

core/AtlasState.cpp
```diff
@@ -128,4 +267,273 @@ namespace atlas
...
+
+        cosim_db_->getTaskQueue()->addWorkerTask(std::move(task));
+
+        if (!all_csr_vals.empty()) {
+            task.reset(new CsrValuesSnapshotter(cosim_db_.get(), hart, pc, all_csr_vals));
+            cosim_db_->getTaskQueue()->addWorkerTask(std::move(task));
+        }
+    }
+
+    uint64_t AtlasState::getMStatusInitialValue(const AtlasState* state, const uint64_t xlen_val)
```
kathlenemagnus, line 533: I'm going to open an issue to replace this with a command line parameter like `--reg mstatus <value>`

core/Fetch.cpp
```diff
@@ -3,6 +3,7 @@
 #include "core/AtlasState.hpp"
 #include "core/Execute.hpp"
 #include "core/Translate.hpp"
+#include "core/Exception.hpp"
```
kathlenemagnus, line 6: Do we need to add these includes? Don't see any changes in this file that need the additional includes.

core/AtlasState.hpp
```diff
@@ -28,6 +33,27 @@ namespace atlas
     class Fetch;
     class Execute;
     class Translate;
+    class Exception;
+
+    class CoSimQuery
```
kathlenemagnus, line 38: Where is the implementation of this class?

core/observers/CoSimObserver.hpp
colby-nyce: I believe this was added to get past some assertion in run_cosim.cpp but that call site was removed. I'll remove the stopSim() stuff added here.

core/AtlasState.cpp
colby-nyce: We need to ask the observer what the previous RD value was (at the beginning of the current inst) to send it to the database. I don't love this part of the code and it's a little flimsy, but the best bet for cleaning it up is to do so when the exception/IDE logging ActionGroup paths are refactored. The IDE observer should be able to do all this by itself. I'm putting together a proposal for the ActionGroup design for Python/C++ co-simulation and that full refactor/proposal would back out this mini hack.

PR #47, opened 2025-01-14
Mavis Extension Manager, Finish ActionGroup and Exception Flow

kathlenemagnus:
Started by integrating the new Mavis ExtensionManager into Atlas and then got carried away fixing other things. Moved Mavis ownership from Fetch to AtlasState to keep the Mavis instance with the ExtensionManager. The ISA string parameter is now used to initialize Mavis and get the XLEN. Future work is still needed to differentiate between "supported" extensions and "enabled" extensions. 

Other things in this PR:
- Created the "finish instruction" ActionGroup to hold common post-execute Actions from AtlasState and the Observers
- Refactored how the Exception unit's ActionGroup is connected
- Added support for a pre-Exception Action in the Observers to capture the trap cause before it is handled
- Added the trap cause to the instruction logger output
- Fixed bug in UID assignment
- Ran clangformat
- Added the new Exception unit to all core tests
- Added a test to the regression with the instruction logger enabled


core/observers/Observer.cpp
```diff
@@ -5,16 +5,27 @@
 namespace atlas
 {
 
-    void Observer::insertExecuteActions(ActionGroup* action_group)
+    void Observer::insertPreExecuteActions(ActionGroup* action_group)
     {
-        if (post_execute_action_)
+        if (pre_execute_action_)
```
colby-nyce, line 10: insertPreExecuteActions, insertPreExceptionActions, and insertFinishActions all have to check enabled() and do nothing if not. I made this change to your branch locally to verify, and it does what I need for the IDE stuff.

colby-nyce (changes requested):
Please check enabled() to avoid the performance hit later.

Edit -- Ignore this comment: I'm going to redesign the InstructionLogger in a quick follow-up submission which will fix this and make the code a little faster.

core/observers/Observer.cpp
colby-nyce: Ignore this comment: I'm going to redesign the InstructionLogger in a quick follow-up submission which will fix this and make the code a little faster.

colby-nyce (approved):

PR #49, opened 2025-01-15
Do not always create the InstructionLogger

colby-nyce:


core/AtlasState.cpp
```diff
@@ -99,12 +94,55 @@ namespace atlas
...
-            "The InstructionLogger should be the only Observer registered at this point!");
-        observers_.back()->insertFinishActions(&finish_action_group_);
-        observers_.back()->insertPreExceptionActions(exception_unit_->getActionGroup());
+        // FIXME: Does Sparta have a callback notif for when debug icount is reached?
+        if (inst_logger_.observed()) {
+            addObserver(std::make_unique<InstructionLogger>(inst_logger_));
+        }
+    }
+
+    ActionGroup* AtlasState::preExecute_(AtlasState* state)
```
kathlenemagnus, line 103: This is a great idea!

core/observers/Observer.hpp
```diff
@@ -44,23 +42,21 @@ namespace atlas
...
-
-        void insertFinishActions(ActionGroup* action_group);
-
         uint64_t getPrevRdValue() const
         {
             sparta_assert(dst_regs_.size() == 1);
             return convertFromByteVector<uint64_t>(dst_regs_[0].reg_prev_value);
         }
 
+        virtual ActionGroup* preExecute(AtlasState*) { return nullptr; }
```
kathlenemagnus, line 51: If we're going to just call these methods directly from `AtlasState::preExecute`, they don't need to return an `ActionGroup`.
```
virtual void preExecute(AtlasState*) {}
```

kathlenemagnus (approved):

PR #52, opened 2025-01-16
New IDE implementation

colby-nyce:


kathlenemagnus-mips:
@colby-nyce what use case have you found where a pre/post execute Observer Action will return a fail Action?

kathlenemagnus (approved):

colby-nyce:
The fail action group code path would be used often by the IDE to tell the C++ sim not to evaluate a particular inst handler.

AtlasIDE.py
```diff
@@ -1,4 +1,4 @@
-import os, sys, wx, sqlite3, re, copy, json
+import os, sys, wx, sqlite3, re, copy, json, socket
```
colby-nyce, line 1: This file will be deleted. It was just a prototype.

core/AtlasState.cpp
```diff
@@ -350,6 +366,51 @@ namespace atlas
         return 0;
     }
 
+    sparta::Register* AtlasState::findRegister(const std::string & reg_name) const
```
kathlenemagnus-mips, line 369: Does the register name here have a prefix like "x" or "f"? If so, this method can probably be a little smarter and use the prefix to determine the register type.

core/AtlasState.cpp
```diff
@@ -350,6 +366,51 @@ namespace atlas
...
+            {
+                continue;
+            }
+            if (reg->getName() == reg_name)
+            {
+                return reg;
+            }
+        }
+
+        return nullptr;
```
kathlenemagnus-mips, line 411: Should this be an error instead? Seems like it would be bad if a register didn't exist at all.

core/AtlasState.cpp
```diff
@@ -576,4 +637,24 @@ namespace atlas
...
+        if (interactive_mode_)
+        {
+            auto observer = std::make_unique<SimController>();
+            sim_controller_ = observer.get();
+            addObserver(std::move(observer));
+            sim_controller_->postInit(this);
+        }
+    }
+
+    void AtlasState::cleanup()
```
kathlenemagnus-mips, line 652: Any reason this can't be included in the destructor?

kathlenemagnus-mips (approved):
I'll be honest I didn't review every single line of Python code in this thing, but everything I did look at looks good!

core/AtlasState.cpp
colby-nyce: Since the RegisterSet's are populated from the AtlasState ctor once, I am caching a map from the reg name to the sparta::Register up front. This will be added to [a separate PR](https://github.com/sparcians/atlas/pull/54).

core/AtlasState.cpp
colby-nyce: I changed the signature to be findRegister(name, must_exist=true) to be similar to many Sparta APIs that do this.

core/AtlasState.cpp
colby-nyce: I added it to the AtlasSim destructor.

PR #59, opened 2025-01-30
First Stab at Supporting Sv32 and Sv39 Translation

kathlenemagnus:
This is very much a WIP. Much of this code was contributed by @skhanx86. I've reworked some of his helper classes (`PageTableEntry` and `PageTable`) and migrated some of his page walking code from the `PageTableWalker` class into the translation unit. I added an Sv32 translation test that does a complete page table walk and translation of a virtual address.


core/translate/Translate.cpp
```diff
@@ -0,0 +1,142 @@
...
+        {
+            translation_state->setTranslationResult(vaddr, request.size);
+            // Keep going
+            return nullptr;
+        }
+
+        // Page size is 4K for both RV32 and RV64
+        constexpr uint64_t PAGESIZE = 4096;
+        uint64_t ppn = READ_CSR_FIELD(SATP, ppn) * PAGESIZE;
+        for (uint32_t level = getNumPageWalkLevels_<Mode>() - 1; level >= 0; --level)
```
colby-nyce, line 99: I'm surprised this didn't warn/error since "level >= 0" is always true, so you could have an infinite loop here.

core/translate/PageTableEntry.hpp
```diff
@@ -0,0 +1,122 @@
...
+            }
+            else
+            {
+                sparta_assert(false, "MMU mode is not currently supported!");
+            }
+
+            // Lower 10 bits are the same for all modes
+            rsw_ = (pte_val_ & translate_types::Sv32::PteFields::rsw.bitmask)
+                   >> translate_types::Sv32::PteFields::rsw.lsb;
+            d_ = pte_val_ & translate_types::Sv32::PteFields::dirty.bitmask;
```
colby-nyce, line 98: The variables like d_ are not good names, but also it doesn't look like you need them at all either since everything is encoded in pte_val_

I would just leave the public methods and do the masking "pte_val_ & ..." in those methods, then remove the d_ etc variables.

core/translate/Translate.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+        std::array<Action, N_MMU_MODES> inst_translation_actions_;
+        std::array<Action, N_MMU_MODES> data_translation_actions_;
+
+        template <MMUMode Mode> uint32_t getNumPageWalkLevels_() const
+        {
+            return num_pagewalk_levels_.at(static_cast<uint32_t>(Mode));
+        }
+
+        static constexpr std::array<uint32_t, N_MMU_MODES> num_pagewalk_levels_{0, 2, 3, 4, 5};
```
colby-nyce, line 52: This looks weird honestly -- I would remove num_pagewalk_levels_ and just "if constexpr (Mode == MMUMode::BAREMETAL) return 0;" etc. since everything is constexpr / templates.

core/translate/Translate.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+        template <MMUMode Mode> uint64_t extractVpn_(const uint32_t level, const uint64_t vaddr)
+        {
+            auto get_vpn_field = [](const uint32_t level) -> const translate_types::FieldDef
```
colby-nyce, line 56: The "const" on the return type doesn't have any effect (since it's returned by value... the const qualifier on return types is for pointers/refs)

core/translate/TranslateTypes.hpp
```diff
@@ -0,0 +1,162 @@
...
+            FieldDef() : bitmask(0xffffffffffffffff), msb(63), lsb(0) {}
+
+            FieldDef(const uint64_t bm, const uint32_t mb, const uint32_t lb) :
+                bitmask(bm),
+                msb(mb),
+                lsb(lb)
+            {
+            }
+
+            FieldDef(const FieldDef & def) : bitmask(def.bitmask), msb(def.msb), lsb(def.lsb) {}
```
colby-nyce, line 25: You can just do:

FieldDef(const FieldDef &rhs) = default;

colby-nyce (approved):

core/translate/Translate.cpp
kathlenemagnus: I'm decrementing level on each iteration, so there's no infinite loop.

core/translate/PageTableEntry.hpp
kathlenemagnus: I agree that makes the class much simpler. I'll do that.

core/translate/Translate.hpp
kathlenemagnus-mips: Fixed.

core/translate/TranslateTypes.hpp
kathlenemagnus: Fixed.

PR #65, opened 2025-02-07
Fixed CSR Instruction Handlers and Use New Register Macros

kathlenemagnus:
Set of fixes needed to get the `rv64mi-p-csr` arch test to pass.
- Use new register macros in all instruction handlers
- Set some CSR read-only fields at boot based on the configured XLEN
- Added a way to override CSR register values through a JSON file
- Made several fixes to the CSR instruction handlers
- Improved Instruction Logger to better support exceptions and to print CSR values for CSR instructions


core/AtlasState.cpp
```diff
@@ -610,16 +641,29 @@ namespace atlas
         }
     }
 
-    uint64_t AtlasState::getMStatusInitialValue(const AtlasState* state, const uint64_t xlen_val)
+    void AtlasState::boot()
     {
-        // TODO cnyce
-        (void)state;
-        (void)xlen_val;
-        return 42949672960;
```
kathlenemagnus, line 618: @colby-nyce FYI the value here (`0xa00000000`) is setting the CSR fields that set the XLEN for user mode (`uxl`) and supervisor mode (`sxl`). The same value is being written to MSTATUS and SSTATUS in  `AtlasState::boot`.

core/AtlasState.cpp
```diff
@@ -152,6 +153,32 @@ namespace atlas
         }
     }
 
+    void AtlasState::onBindTreeLate_()
+    {
+        // Write initial values to CSR registers
```
kathlenemagnus, line 158: I'm not committed to this method of providing CSR values. Open to suggestions. I've used complex, Sparta parameters in the past, but I've found that they can be hard to read in the yaml file sometimes.

arch/default_csr_values.json
```diff
@@ -0,0 +1,2 @@
+{
```
kathlenemagnus, line 1: Format would be like this:
```
{
    "mstatus": "0xa00000000",
    "sstatus": "0x200000000"
}
```

core/AtlasState.cpp
```diff
@@ -525,11 +556,11 @@ namespace atlas
...
-        std::string rs1_name = insn->hasRs1() ? insn->getRs1()->getName() : "";
-        uint64_t rs1_val = insn->hasRs1() ? insn->getRs1()->dmiRead<uint64_t>() : 0;
-        std::string rs2_name = insn->hasRs2() ? insn->getRs2()->getName() : "";
-        uint64_t rs2_val = insn->hasRs2() ? insn->getRs2()->dmiRead<uint64_t>() : 0;
-        std::string rd_name = insn->hasRd() ? insn->getRd()->getName() : "";
+        std::string rs1_name = insn->hasRs1() ? insn->getRs1Reg()->getName() : "";
+        uint64_t rs1_val = insn->hasRs1() ? insn->getRs1Reg()->dmiRead<uint64_t>() : 0;
+        std::string rs2_name = insn->hasRs2() ? insn->getRs2Reg()->getName() : "";
+        uint64_t rs2_val = insn->hasRs2() ? insn->getRs2Reg()->dmiRead<uint64_t>() : 0;
+        std::string rd_name = insn->hasRd() ? insn->getRdReg()->getName() : "";
```
knute-mips, line 563: Could be `const`

core/AtlasInst.hpp
```diff
@@ -56,22 +67,40 @@ namespace atlas
...
+
+        sparta::Register* getRs2Reg()
+        {
+            sparta_assert(rs2_reg_, "Operand RS2 is a nullptr! " << *this);
+            return rs2_reg_;
+        }
+
+        sparta::Register* getRdReg()
+        {
+            sparta_assert(rd_reg_, "Operand RD is a nullptr! " << *this);
```
knute-mips, line 102: Wondering if you should use `assert` to speed this up for release builds

knute-mips (approved):

core/AtlasState.cpp
kathlenemagnus: Fixed.

core/AtlasInst.hpp
```diff
@@ -56,22 +67,40 @@ namespace atlas
...
+            return rs2_->field_value;
         }
 
-        sparta::Register* getRd()
+        uint32_t getRd()
         {
             sparta_assert(rd_, "Operand RD is a nullptr! " << *this);
-            return rd_;
+            return rd_->field_value;
+        }
```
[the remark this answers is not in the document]
kathlenemagnus: Fixed.

core/AtlasInst.hpp
kathlenemagnus: Good idea for the future, but I'd rather leave them in until Atlas is semi-functional.

core/AtlasInst.hpp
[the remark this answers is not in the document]
knute-mips: Your 🍺 tab is getting quite big. 🤣

arch/register_macros.hpp
```diff
@@ -72,3 +71,18 @@
...
+#define POKE_CSR_FIELD(reg_ident, field_name, field_value)                                         \
+    {                                                                                              \
+        auto csr_value = state->getCsrRegister(reg_ident)->dmiRead<uint64_t>();                    \
+                                                                                                   \
+        const auto low_bit = atlas::getCsrBitRange(reg_ident, #field_name).first;                  \
+        const auto high_bit = atlas::getCsrBitRange(reg_ident, #field_name).second;                \
+        const auto mask = ((1ULL << (high_bit - low_bit + 1)) - 1) << low_bit;                     \
+        csr_value &= ~mask;                                                                        \
+                                                                                                   \
+        const uint64_t new_field_value = field_value << low_bit;                                   \
```
colby-nyce, line 84: You need to cast field_value to uint64_t before the shift, otherwise you'll get the same compiler issue if the macro was called with a sub-64-bit field_value type.

core/AtlasInst.hpp
```diff
@@ -102,9 +131,12 @@ namespace atlas
         Addr next_pc_;
 
         // Registers
-        sparta::Register* rs1_;
-        sparta::Register* rs2_;
-        sparta::Register* rd_;
+        const mavis::OperandInfo::Element* rs1_;
```
colby-nyce, line 134: Can you rename these to "rs1_info_" or "rs1_elem_"?

core/AtlasState.cpp
```diff
@@ -597,29 +627,42 @@ namespace atlas
...
-        (void)xlen_val;
-        return 42949672960;
-    }
+        std::cout << "Booting hartid " << std::dec << hart_id_ << std::endl;
+        {
+            AtlasState* state = this;
+
+            POKE_CSR_REG(MHARTID, hart_id_);
+
+            // TODO: Initiali MISA CSR with XLEN and enabled extensions
```
colby-nyce, line 651: Initialize

core/inst_handlers/zicsr/RvzicsrInsts.cpp
```diff
@@ -17,193 +17,226 @@ namespace atlas
...
+        auto csr_reg = state->getCsrRegister(csr);
+        if (!csr_reg)
         {
             THROW_ILLEGAL_INSTRUCTION;
         }
 
-        const reg_t old = reg->dmiRead<uint64_t>();
-
-        if (write)
+        const XLEN csr_val = READ_CSR_REG(csr);
```
colby-nyce, line 91: I think we should put the THROW_ILLEGAL_INSTRUCTION inside READ_CSR_REG / WRITE_CSR_REG etc including POKE and *_FIELD methods.

core/inst_handlers/zicsr/RvzicsrInsts.cpp
```diff
@@ -17,193 +17,226 @@ namespace atlas
...
-        const reg_t old = reg->dmiRead<uint64_t>();
-        const uint64_t imm = insn->getSignExtendedImmediate<RV64, 5>();
-        if (imm)
+        const XLEN rs1_val = READ_INT_REG(rs1);
+        // Only read CSR if rd!=x0
+        if (rd != 0)
         {
-            reg->write(old | imm); // dmiWrite?
+            const XLEN csr_val = zext(READ_CSR_REG(csr), state->getXlen());
+            // TODO: Use macro to avoid writing to x0
```
colby-nyce, line 206: What does this comment mean? The underlying write()/dmiWrite() method is already protected against writes to X0.

core/inst_handlers/zicsr/RvzicsrInsts.cpp
```diff
@@ -17,193 +17,226 @@ namespace atlas
...
         return nullptr;
     }
 
-    ActionGroup* RvzicsrInsts::csrrci_64_handler(atlas::AtlasState* state)
+    template <typename XLEN> ActionGroup* RvzicsrInsts::csrrwi_handler(atlas::AtlasState* state)
     {
         const AtlasInstPtr & insn = state->getCurrentInst();
 
+        const XLEN imm = insn->getImmediate();
+        const auto & rd = insn->getRd();
```
colby-nyce, line 220: I'm not sure how this compiles since the api is "uint64_t getRd() const" and you cannot take a reference to a temp value.

colby-nyce: Edit: You cannot take a non-const ref to an rvalue, so in this case I guess it compiles down to the same thing. It still looks odd IMO and I'd prefer "const auto rd" to match the API returning a value type.

arch/register_macros.hpp
kathlenemagnus-mips: Fixed.

core/AtlasInst.hpp
kathlenemagnus-mips: I changed it to `rs1_operand_`.

core/inst_handlers/zicsr/RvzicsrInsts.cpp
kathlenemagnus-mips: I actually was thinking of putting it in Decode. Architecturally, that is where the exception would really be thrown.

core/inst_handlers/zicsr/RvzicsrInsts.cpp
kathlenemagnus-mips: The comment is wrong and no longer relevant. Removing.

core/inst_handlers/zicsr/RvzicsrInsts.cpp
kathlenemagnus-mips: The ampersand was left over from when `getRd()` returned a Sparta register pointer. It doesn't make sense anymore. I've removed it throughout this file.

core/AtlasState.cpp
kathlenemagnus: Fixed.

core/inst_handlers/zicsr/RvzicsrInsts.cpp
```diff
@@ -17,193 +17,226 @@ namespace atlas
...
+        if (!csr_reg)
         {
             THROW_ILLEGAL_INSTRUCTION;
         }
 
-        const reg_t old = reg->dmiRead<uint64_t>();
-
-        if (write)
+        const XLEN csr_val = READ_CSR_REG(csr);
+        // Don't wqrite CSR is rs1=x0
```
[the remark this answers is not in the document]
kathlenemagnus: Fixed.

core/inst_handlers/zicsr/RvzicsrInsts.cpp
```diff
@@ -17,193 +17,226 @@ namespace atlas
...
         {
-            reg->write(old | rs1_val); // dmiWrite?
+            // rs1 value is treated as a bit mask to clear bits
+            const XLEN rs1_val = READ_INT_REG(rs1);
+            WRITE_CSR_REG(csr, (~rs1_val & csr_val));
         }
 
-        const auto rd_val = sext(old, state->getXlen());
-        insn->getRd()->write(rd_val); // dmiWrite?
+        WRITE_INT_REG(rd, csr_val);
```
lzhu-mips, line 94: I am wondering how rd==x0 is taken care of.  I believe if rd==x0, no value is written. Is that checked in WRITE_INT_REG?

kathlenemagnus: Yes, it is taken care of in the macro. If `reg_ident == 0`, then the write does not occur.

lzhu-mips (approved):

lzhu-pub (approved):

PR #68, opened 2025-02-13
Proposal for XLEN-Aware Register Macros

kathlenemagnus:


arch/register_macros.hpp
knute-mips, line 1: Doesn't this kinda warrant inline templates...?   😢

knute-mips (approved):

arch/register_macros.hpp
lzhu-mips: Exactly. seems there is no benefit using macro here. But for consistency....

arch/register_macros.hpp
```diff
@@ -11,22 +11,41 @@
...
+
+#define WRITE_INT_REG32(reg_ident, reg_value)                                                      \
+    if (reg_ident != 0)                                                                            \
+        state->getIntRegister(reg_ident)->dmiWrite<uint32_t>(reg_value);
+
+#define WRITE_INT_REG64(reg_ident, reg_value)                                                      \
     if (reg_ident != 0)                                                                            \
         state->getIntRegister(reg_ident)->dmiWrite<uint64_t>(reg_value);
 
-#define READ_FP_FIELD(reg_ident, field_name)                                                       \
```
colby-nyce, line 21: Is reading and writing FP fields no longer supported?

colby-nyce:
I think this PR should be discussed as this seems to be going in the wrong direction (reusable APIs to hard-coded). It's not clear what the benefit of this is, but I'm envisioning more and more near-duplicate handler code to come out of this over time.

arch/register_macros.hpp
kathlenemagnus: @lzhu-pub and I talked about it and determined it's not necessary since the softfloat functions take care of handling the upper bits of the 64-bit register when doing 32-bit single precision operations. We want to leave the fields defined though, because it's nice for debugging.

colby-nyce (approved):

kathlenemagnus:
@knute-mips @colby-nyce @lzhu-pub I've just pushed an updated proposal that should meet everyone requirements. I've moved the register macro definitions to `AtlasState.hpp` and they are now static, inlined functions. The first parameter is a pointer to `AtlasState`. The functions are also templated with the XLEN. The template parameter is expected to be used like this:
```
template <typename XLEN>
ActionGroup* example_inst_handler(AtlasState* state)
{
        
        // XLEN-aware register read                
        const XLEN rs1_val = READ_INT_REG<XLEN>(state, insn->getRs1());

        // Explicit 64-bit read
        const uint64_t rs1_val64 = READ_INT_REG<uint64_t>(state, insn->getRs1());

        // Explicit 32-bit read
        const uint32_t rs1_val32 = READ_INT_REG<uint32_t>(state, insn->getRs1());
}
```

I'm not committed to the names of the functions. Open to suggestions there.

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+        }
+        else
+        {
+            state->getCsrRegister(reg_ident)->dmiWrite(reg_value);
+        }
+    }
+
+    template <typename XLEN> static inline XLEN PEEK_CSR_REG(AtlasState* state, uint32_t reg_ident)
+    {
+        READ_CSR_REG<XLEN>(state, reg_ident);
```
colby-nyce, line 401: Need to return the value :)

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+    }
+
+    template <typename XLEN>
+    static inline void WRITE_CSR_FIELD(AtlasState* state, uint32_t reg_ident,
+                                       const char* field_name, XLEN field_value)
+    {
+        XLEN csr_value = READ_CSR_REG<XLEN>(state, reg_ident);
+
+        const XLEN field_lsb = atlas::getCsrBitRange(reg_ident, field_name).first;
+        const XLEN field_msb = atlas::getCsrBitRange(reg_ident, field_name).second;
```
colby-nyce, line 426: Does the generated getCsrBitRange() method need to be split into 32- and 64-bit versions? Because this code is returning ranges using 0-63 which get truncated to XLEN for rv32. Not that it prevents this submission - but it's something to keep in mind.

test/register/macros/RegisterMacros_test.cpp
```diff
@@ -254,7 +256,7 @@ void testCsrRegs()
 int main()
 {
     testIntRegs();
-    testFpRegs();
+    // testFpRegs();
```
colby-nyce, line 259: What was the issue here?

core/AtlasState.hpp
kathlenemagnus: Good catch! Fixed,

core/AtlasState.hpp
kathlenemagnus: Yes it does need to be split. Is this file auto-generated for RV64 only?

test/register/macros/RegisterMacros_test.cpp
kathlenemagnus: I just need to update the test since the macros changed.

test/register/CMakeLists.txt
```diff
@@ -7,4 +7,4 @@ atlas_named_test(Register_test_run AtlasRegister_test)
 atlas_copy(AtlasRegister_test *.json)
 atlas_copy(AtlasRegister_test ../../arch/rv64/reg_*.json)
 
-add_subdirectory(macros)
+#add_subdirectory(macros)
```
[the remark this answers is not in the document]
kathlenemagnus: I'll need to rewrite this tester to use the new macros. I'll work on it this afternoon.

core/AtlasState.cpp
```diff
@@ -646,16 +645,16 @@ namespace atlas
         {
             AtlasState* state = this;
 
-            POKE_CSR_REG(MHARTID, hart_id_);
+            POKE_CSR_REG(this, MHARTID, hart_id_);
 
             // TODO: Initialize MISA CSR with XLEN and enabled extensions
             const uint64_t xlen_val = (xlen_ == 64) ? 2 : 1;
-            POKE_CSR_FIELD(MISA, mxl, xlen_val);
+            POKE_CSR_FIELD(this, MISA, "mxl", xlen_val);
```
colby-nyce, line 652: I see a problem with POKE_CSR_FIELD, which may apply to others. The XLEN under the hood is determined solely by the data type of the fourth arg (xlen_val). So with "uint64_t xlen_val", we will never hit the getCsrBitRange<uint32_t> specialization. I suspect that for XLEN=32, all this code blows up (the bit ranges are going up to 0-63 even though rv32 cannot go that high, so there would be wrong answers of all kinds).

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+        state->getVecRegister(reg_ident)->dmiWrite<XLEN>(reg_value);
+    }
+
+    template <typename XLEN> static inline XLEN READ_CSR_REG(AtlasState* state, uint32_t reg_ident)
+    {
+        return state->getCsrRegister(reg_ident)->dmiRead<XLEN>();
+    }
+
+    template <typename XLEN>
+    static inline void WRITE_CSR_REG(AtlasState* state, uint32_t reg_ident, XLEN reg_value)
```
colby-nyce, line 383: I think we have to use "uint64_t reg_value" instead of "XLEN reg_value" to force the call sites to say WRITE_CSR_REG<XLEN>(...)

Otherwise I see a lot of bugs happening around accidental use of 64-bit template specializations when using rv32.

uint64_t reg_value = 0xff;
WRITE_CSR_REG(state, 7, reg_value);

^^^ Despite being rv32, that code will call getCsrBitMask<uint64_t> instead of getCsrBitMask<uint32_t> and the mask will be all wrong.

core/AtlasState.hpp
colby-nyce: Edit -- yes it does (GenCSRHeaders.py)

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+        state->getCsrRegister(reg_ident)->dmiWrite(reg_value);
+    }
+
+    template <typename XLEN>
+    static inline XLEN READ_CSR_FIELD(AtlasState* state, uint32_t reg_ident, const char* field_name)
+    {
+        const XLEN field_lsb = atlas::getCsrBitRange(reg_ident, field_name).first;
+        const XLEN field_msb = atlas::getCsrBitRange(reg_ident, field_name).second;
+        return ((state->getCsrRegister(reg_ident)->dmiRead<uint64_t>() >> field_lsb)
+                & ((1ULL << (field_msb - field_lsb + 1)) - 1));
```
lzhu-mips, line 416: what if msb == 63 and lsb == 0? then the shift would be 64. It might give you a compile warning.

core/AtlasState.cpp
kathlenemagnus-mips: Good suggestion. I will use `uint64_t` for the value param so that the template parameter has to be explicitly provided.

core/AtlasState.hpp
lzhu-mips: template <typename XLEN>
static inline void WRITE_CSR_REG(AtlasState* state, uint32_t reg_ident, typename std::enable_if<true, const XLEN>::type reg_value)
{
    ...
}

this will force XLEN to be explicitly provided.

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+
+    template <typename XLEN> static inline XLEN READ_INT_REG(AtlasState* state, uint32_t reg_ident)
+    {
+        return (reg_ident == 0) ? 0 : state->getIntRegister(reg_ident)->dmiRead<XLEN>();
+    }
+
+    template <typename XLEN>
+    static inline void WRITE_INT_REG(AtlasState* state, uint32_t reg_ident, XLEN reg_value)
+    {
+        if (reg_ident != 0)
```
colby-nyce, line 349: Do you recall that I added this isWritable() check inside write() and dmiWrite()? (In the Sparta code)

If we want to do the checks here, should we remove isWritable() from Sparta for the performance boost?

If we leave it in Sparta, then I think we shouldn't check for X0 anywhere in Atlas.

core/AtlasState.hpp
colby-nyce: We can keep it simpler with just this:

```
template < typename XLEN >
static inline void WRITE_CSR_REG(AtlasState*, uint32_t, uint64_t reg_value)
{...}
```

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+
+    template <typename XLEN> static inline XLEN READ_CSR_REG(AtlasState* state, uint32_t reg_ident)
+    {
+        return state->getCsrRegister(reg_ident)->dmiRead<XLEN>();
+    }
+
+    template <typename XLEN>
+    static inline void WRITE_CSR_REG(AtlasState* state, uint32_t reg_ident, uint64_t reg_value)
+    {
+        if (atlas::getCsrBitMask<XLEN>(reg_ident) != 0xffffffffffffffff)
```
colby-nyce, line 385: We can't hard code `0xffffffffffffffff` so use `std::numeric_limits<XLEN>::max()` instead.

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+        {
+            auto reg = state->getCsrRegister(reg_ident);
+            const auto old_value = reg->dmiRead<uint64_t>();
```
colby-nyce, line 388: This has to be <XLEN>. As it is, it upcasts everything to uint64_t and dmiWrite<uint32_t> is never called on line 391.

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+                & ((1ULL << (field_msb - field_lsb + 1)) - 1));
+    }
+
+    template <typename XLEN>
+    static inline void WRITE_CSR_FIELD(AtlasState* state, uint32_t reg_ident,
+                                       const char* field_name, XLEN field_value)
+    {
+        XLEN csr_value = READ_CSR_REG<XLEN>(state, reg_ident);
+
+        const XLEN field_lsb = atlas::getCsrBitRange(reg_ident, field_name).first;
```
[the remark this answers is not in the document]
kathlenemagnus-mips: Fixed.

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+    {
+        state->getCsrRegister(reg_ident)->dmiWrite(reg_value);
+    }
+
+    template <typename XLEN>
+    static inline XLEN READ_CSR_FIELD(AtlasState* state, uint32_t reg_ident, const char* field_name)
+    {
+        const XLEN field_lsb = atlas::getCsrBitRange(reg_ident, field_name).first;
+        const XLEN field_msb = atlas::getCsrBitRange(reg_ident, field_name).second;
+        return ((state->getCsrRegister(reg_ident)->dmiRead<uint64_t>() >> field_lsb)
```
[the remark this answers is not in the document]
kathlenemagnus-mips: Fixed.

core/AtlasState.hpp
kathlenemagnus-mips: What do you think is the best choice for simulator performance? I am ok with either solution.

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+        }
+        else
+        {
+            state->getCsrRegister(reg_ident)->dmiWrite(reg_value);
+        }
+    }
+
+    template <typename XLEN> static inline XLEN PEEK_CSR_REG(AtlasState* state, uint32_t reg_ident)
+    {
+        READ_CSR_REG<XLEN>(state, reg_ident);
```
colby-nyce, line 401: Still need to return `READ_CSR_REG`

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+    }
+
+    template <typename XLEN>
+    static inline void POKE_CSR_REG(AtlasState* state, uint32_t reg_ident, uint64_t reg_value)
+    {
+        state->getCsrRegister(reg_ident)->dmiWrite(reg_value);
```
colby-nyce, line 407: This needs to be `dmiWrite<XLEN>(reg_value)`

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+    template <typename XLEN>
+    static inline void WRITE_CSR_REG(AtlasState* state, uint32_t reg_ident, uint64_t reg_value)
+    {
+        if (atlas::getCsrBitMask<XLEN>(reg_ident) != 0xffffffffffffffff)
+        {
+            auto reg = state->getCsrRegister(reg_ident);
+            const auto old_value = reg->dmiRead<uint64_t>();
+            const auto mask = atlas::getCsrBitMask<XLEN>(reg_ident);
+            const auto write_val = (old_value & ~mask) | (reg_value & mask);
+            reg->dmiWrite(write_val);
```
colby-nyce, line 391: `dmiWrite<XLEN>(write_val)`

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+        }
+        else
+        {
+            state->getCsrRegister(reg_ident)->dmiWrite(reg_value);
```
colby-nyce, line 395: `dmiWrite<XLEN>(reg_value)`

core/AtlasState.hpp
kathlenemagnus-mips: Fixed.

core/AtlasState.hpp
```diff
@@ -333,4 +337,116 @@ namespace atlas
...
+    template <typename XLEN>
+    static inline void POKE_CSR_REG(AtlasState* state, uint32_t reg_ident, uint64_t reg_value)
+    {
+        state->getCsrRegister(reg_ident)->dmiWrite(reg_value);
+    }
+
+    template <typename XLEN>
+    static inline XLEN READ_CSR_FIELD(AtlasState* state, uint32_t reg_ident, const char* field_name)
+    {
+        const XLEN field_lsb = atlas::getCsrBitRange<XLEN>(reg_ident, field_name).first;
```
colby-nyce, line 413: Please only call getCsrBitRange() once here to avoid an unnecessary map lookup.

core/AtlasState.hpp
kathlenemagnus: I added in the updated Python script you sent me and updated these function calls to specify the new XLEN template param.

test/register/CMakeLists.txt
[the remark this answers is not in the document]
kathlenemagnus: Tester has been updated and is passing.

core/AtlasState.hpp
kathlenemagnus: Fixed.

core/AtlasState.hpp
kathlenemagnus: Fixed.

core/AtlasState.hpp
kathlenemagnus: Fixed.

core/AtlasState.hpp
kathlenemagnus: Fixed.

core/AtlasState.hpp
kathlenemagnus: Fixed to only call `getCsrBitRang()` once for all register functions.

kathlenemagnus:
@colby-nyce thank you for the VERY thorough review of this code! I really like where we ended up with this. Regression is passing with the register macro test added back in and I confirmed that the pass rate of the RISC-V arch tests is unchanged (91/130).

core/AtlasState.hpp
colby-nyce: The compiler won't issue a warning about that, since it can't know the runtime values (note that they were compile-time values when these were a macro, which is why we hit the compiler error before).

BUT it is still undefined behavior to do `1ULL << 64` -- Kathlene, can you break apart lines 415-416 to take this into account? If the field spans the whole register then no mask/shift is needed. But you'll need code that works for both XLENs.

core/AtlasState.hpp
colby-nyce: I think it's better to remove `if (!isWritable()) return;` from Sparta and maybe replace it with `assert(isWritable());` (not sparta_assert)

The issue with Sparta's implementation is that every register is going to check against X0, even if we are talking to an FP/VEC/CSR. I'll change that in a separate PR for Sparta, but for Atlas just leave the checks for INT regs only before writing the value.

core/AtlasState.hpp
```diff
@@ -333,4 +337,131 @@ namespace atlas
...
+    {
+        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
+        state->getCsrRegister(reg_ident)->dmiWrite<XLEN>(reg_value);
+    }
+
+    template <typename XLEN>
+    static inline XLEN READ_CSR_FIELD(AtlasState* state, uint32_t reg_ident, const char* field_name)
+    {
+        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
+        const auto csr_bit_range = atlas::getCsrBitRange<XLEN>(reg_ident, field_name);
```
colby-nyce, line 423: Note that getCsrBitRange() returns a reference, so you should do `const auto &`

core/AtlasState.hpp
```diff
@@ -333,4 +337,131 @@ namespace atlas
...
+    }
+
+    template <typename XLEN>
+    static inline void WRITE_CSR_FIELD(AtlasState* state, uint32_t reg_ident,
+                                       const char* field_name, uint64_t field_value)
+    {
+        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
+        XLEN csr_value = READ_CSR_REG<XLEN>(state, reg_ident);
+
+        const auto csr_bit_range = atlas::getCsrBitRange<XLEN>(reg_ident, field_name);
```
colby-nyce, line 437: `const auto &`

core/AtlasState.hpp
```diff
@@ -333,4 +337,131 @@ namespace atlas
...
+        const XLEN field_lsb = csr_bit_range.first;
+        const XLEN field_msb = csr_bit_range.second;
+        const XLEN mask = ((1ULL << (field_msb - field_lsb + 1)) - 1) << field_lsb;
```
colby-nyce, line 440: Instead of `1ULL1` use `XLEN(1)`

core/AtlasState.hpp
```diff
@@ -333,4 +337,131 @@ namespace atlas
...
+    static inline void POKE_CSR_FIELD(AtlasState* state, uint32_t reg_ident, const char* field_name,
+                                      uint64_t field_value)
+    {
+        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
+        XLEN csr_value = READ_CSR_REG<XLEN>(state, reg_ident);
+
+        const auto csr_bit_range = atlas::getCsrBitRange<XLEN>(reg_ident, field_name);
+        const XLEN field_lsb = csr_bit_range.first;
+        const XLEN field_msb = csr_bit_range.second;
+        const XLEN mask = ((1ULL << (field_msb - field_lsb + 1)) - 1) << field_lsb;
```
colby-nyce, line 459: Instead of `1ULL` use `XLEN(1)`

core/AtlasState.hpp
```diff
@@ -333,4 +337,131 @@ namespace atlas
         std::unordered_map<std::string, int> reg_ids_by_name_;
         SimController* sim_controller_ = nullptr;
     };
+
+    template <typename XLEN> static inline XLEN READ_INT_REG(AtlasState* state, uint32_t reg_ident)
+    {
+        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
```
colby-nyce, line 343: While I am making the `isWritable()` change in Sparta, I will add this static_assert to the template write() and dmiWrite() methods. You can remove all of them from Atlas as they'll be in Sparta soon.

kathlenemagnus-mips: I don't think this `static assert` makes sense in Sparta. Do you mean you want to only allow 32-bit writes for 32-bit registers and 64-bit writes for 64-bit registers? I don't think the Register class should be that restrictive.

core/AtlasState.hpp
kathlenemagnus: Fixed.

core/AtlasState.hpp
kathlenemagnus: Fixed.

core/AtlasState.hpp
kathlenemagnus: Fixed.

core/AtlasState.hpp
kathlenemagnus: Fixed.

core/AtlasState.hpp
colby-nyce: You are correct, it doesn't belong in Sparta. The static_asserts are all okay here.

core/AtlasState.hpp
```diff
@@ -333,4 +337,131 @@ namespace atlas
...
+
+    template <typename XLEN>
+    static inline XLEN READ_CSR_FIELD(AtlasState* state, uint32_t reg_ident, const char* field_name)
+    {
+        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
+        const auto & csr_bit_range = atlas::getCsrBitRange<XLEN>(reg_ident, field_name);
+        const XLEN field_lsb = csr_bit_range.first;
+        const XLEN field_msb = csr_bit_range.second;
+        return ((state->getCsrRegister(reg_ident)->dmiRead<XLEN>() >> field_lsb)
+                & ((XLEN(1) << (field_msb - field_lsb + 1)) - 1));
```
colby-nyce, line 427: The only remaining issue I see in this PR is what Liya pointed out about the <<64 undefined behavior. I'll approve the PR when that gets updated.

kathlenemagnus: I add an if-statement to check if the CSR field takes up the entire register (0->63 for RV64, 0->31 for RV32) for the read, write and poke CSR field functions.

PR #73, opened 2025-02-20
updating f/d extensions based on recent main change

lzhu-pub:


core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        uint64_t static_rm = inst->getRM();
+        if (static_rm == 7) // RM field "DYN"
+        {
+            static_rm = READ_CSR_REG<XLEN>(state, FRM);
+        }
+        return static_rm;
+    }
+
+    template <typename XLEN> inline void update_csr(AtlasState* state)
```
kathlenemagnus, line 29: Now that I see this code, I am OK with this implementation. Each instruction handler can call this function directly.

I would prefer that all of these functions were private methods of `RvdInsts` instead of being in a separate header file since they're only being called by the D instruction handlers. If the F instruction handlers need to call them as well, you can make them static OR put them in a common base class.

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        // FFLAGS
+        WRITE_CSR_FIELD<XLEN>(
+            state, FFLAGS, "NX",
+            static_cast<uint64_t>((softfloat_exceptionFlags & softfloat_flag_inexact) != 0));
```
kathlenemagnus, line 34: Are `softfloat_exceptionFlags` and `softfloat_flag_inexact` globals of the softfloat library? Wondering how we will be able to support multi-thread/core simulations in the future...

core/inst_handlers/f/CMakeLists.txt
```diff
@@ -5,3 +5,5 @@ add_library(rvf
     OBJECT
     ${SOURCES}
 )
+
```
kathlenemagnus, line 8: Remove extra whitespaces.

kathlenemagnus:
Other than moving the functions in `inst_helpers.hpp` into the instruction handler classes, this looks good to me. IMO the writes to the FFLAGS and FCSR registers as part of the instructions and not a side effects. If there are side effects to writing the FFLAGS and FCSR, then we may want to consider putting that into a post-execute Action.

kathlenemagnus (approved):

core/inst_handlers/f/inst_helpers.hpp
lzhu-pub: yes. they are globals from softfloat. For multi=thread, when thread switching. I presume a restore/save operation for the variable is needed.

core/inst_handlers/f/inst_helpers.hpp
lzhu-pub: yes. the f/d extensions both call updata_csr(). When you say to mark this function as static, are you tying to avoid the multiple copies of "inline"; is that the concern you have? Or you prefer me to define some interface class for both RvfInsts and RvdInsts so we can get rid of the inst_helpsers.h at all?

lzhu-pub:
> Other than moving the functions in `inst_helpers.hpp` into the instruction handler classes, this looks good to me. IMO the writes to the FFLAGS and FCSR registers as part of the instructions and not a side effects. If there are side effects to writing the FFLAGS and FCSR, then we may want to consider putting that into a post-execute Action.

lzhu-pub:
Writing to FFLAGS csr does have side effect. It shall update the softfloat global variable "softfloat_exceptionFlags".

core/inst_handlers/f/CMakeLists.txt
lzhu-pub: solved.

knute-mips:
Curious -- why did you close this PR?

PR #76, opened 2025-02-27
Implemented UART and Added Placeholder for Future Magic Memory Support

kathlenemagnus:
Added support for installing UART and Magic Memory devices in memory. Provided parameters to enabling the UART and setting its base address and size. Magic memory is enabled automatically if "tohost" and "fromhost" symbols are found in the ELF.

I did not provide an implementation of Magic Memory in this PR. That will be done later.


system/MagicMemory.cpp
```diff
@@ -0,0 +1,51 @@
...
+    bool MagicMemory::tryRead_(sparta::memory::addr_t addr, sparta::memory::addr_t size,
+                               uint8_t* buf, const void*, void*)
+    {
+        (void)addr;
+        (void)size;
+        // TODO: Implement magic memory reads
+        return true;
+    }
+
+    bool MagicMemory::tryWrite_(sparta::memory::addr_t addr, sparta::memory::addr_t size,
```
kathlenemagnus, line 25: @ayeshanh this is where we will implement ending the test on a magic memory write. We will inspect the value being written to determine what action should be taken.

knute-mips (approved):
Neat

system/AtlasSystem.cpp
```diff
@@ -12,6 +12,15 @@ namespace atlas
         const std::string workload = p->workload;
         loadWorkload_(workload);
 
+        if (p->enable_uart)
+        {
+            sparta::ResourceTreeNode* uart_rtn = nullptr;
+            tree_nodes_.emplace_back(uart_rtn = new sparta::ResourceTreeNode(
+                                         sys_node, "uart", "SimpleUART", &uart_fact_));
+            uart_rtn->finalize();
+            uart_ = uart_rtn->getResourceAs<SimpleUART>();
```
colby-nyce, line 21: It's a little out-of-model to call finalize() like this manually. It would be more like the other Atlas code if you override onBindTreeEarly_() and set uart_ there without finalize() here.

system/AtlasSystem.cpp
```diff
@@ -48,6 +57,66 @@ namespace atlas
...
+                symbols.get_symbol(symbol_id, name, addr, size, bind, type, section, other);
+
+                // Intentionally overwriting symbols with the same address
+                if (name.empty() == false)
+                {
+                    symbols_[addr] = name;
+
+                    if (name == "tohost")
+                    {
+                        tohost_addr = addr;
```
colby-nyce, line 84: I think this needs a comment. Why are we overwriting to/fromhost_addr in the loop and just quietly taking the last one?

system/AtlasSystem.cpp
```diff
@@ -48,6 +57,66 @@ namespace atlas
...
+        }
+
+        // Magic Memory
+        if (tohost_addr.isValid() && fromhost_addr.isValid())
+        {
+            sparta::memory::addr_t base_addr =
+                std::min(tohost_addr.getValue(), fromhost_addr.getValue());
+            const sparta::memory::addr_t addr_delt =
+                std::max(tohost_addr.getValue() - base_addr, fromhost_addr.getValue() - base_addr);
+            const sparta::memory::addr_t size_minimum = addr_delt + 8;
```
colby-nyce, line 101: I don't care for hardcoded numbers like this "8". Can this be sizeof(uint64_t) or whatever is most appropriate for self-documenting code?

system/AtlasSystem.cpp
```diff
@@ -77,50 +146,126 @@ namespace atlas
...
+        MemObj* mem_obj = nullptr;
+
+        // The allocated memory blocks (Magic Mem, UART, etc)
+        struct AllocatedMemoryBlock
         {
-            ELFIO::symbol_section_accessor symbols(elf_reader_, section.get());
-            for (uint32_t symbol_id = 0; symbol_id < symbols.get_symbols_num(); ++symbol_id)
+            AllocatedMemoryBlock(sparta::memory::addr_t _start_address,
+                                 sparta::memory::addr_t _size) :
+                start_address(_start_address),
```
colby-nyce, line 170: FYI -- it looks like you're using the underscore (_start_address) to differentiate between the member var / ctor list and the ctor args. Modern compilers won't complain:
```
struct Foo {
  Foo(int x) : x(x) {}
  int x;
};
```

system/AtlasSystem.cpp
```diff
@@ -77,50 +146,126 @@ namespace atlas
...
+
+        ////////////////////////////////////////////////////////////////////////////////
+        // Now fill in the memory "blanks"
+        sparta::memory::addr_t addr_block_start = 0;
+        const uint32_t illop = 0;
+        uint32_t block_num = 1;
+
+        while (false == allocated_blocks.empty())
+        {
+            const auto alloc_block = *(allocated_blocks.begin());
```
colby-nyce, line 230: const auto &

system/AtlasSystem.cpp
```diff
@@ -77,50 +146,126 @@ namespace atlas
...
+                                                    ATLAS_SYSTEM_TOTAL_MEMORY - addr_block_start,
+                                                    illop, sizeof(illop)));
+        tree_nodes_.emplace_back(
+            memory_if =
+                new BMOIfNode(sys_node, "mb_" + std::to_string(block_num),
+                              sparta::TreeNode::GROUP_NAME_NONE, sparta::TreeNode::GROUP_IDX_NONE,
+                              "mb_" + std::to_string(block_num), nullptr, *mem_obj));
+        memory_map_->addMapping(addr_block_start, ATLAS_SYSTEM_TOTAL_MEMORY, memory_if,
+                                0x0 /* Additional offset */);
+        memory_map_->dumpMappings(std::cout);
```
colby-nyce, line 269: How verbose is this? Should it be a command line option?

system/MagicMemory.cpp
```diff
@@ -0,0 +1,51 @@
...
+        size_(params->size)
+    {
+    }
+
+    bool MagicMemory::tryRead_(sparta::memory::addr_t addr, sparta::memory::addr_t size,
+                               uint8_t* buf, const void*, void*)
+    {
+        (void)addr;
+        (void)size;
+        // TODO: Implement magic memory reads
```
colby-nyce, line 21: Might as well (void)buf

system/MagicMemory.cpp
```diff
@@ -0,0 +1,51 @@
...
+        // Peeks are allowed, but have no side effects
+        return true;
+    }
+
+    bool MagicMemory::tryPoke_(sparta::memory::addr_t addr, sparta::memory::addr_t size,
+                               const uint8_t* buf)
+    {
+        (void)addr;
+        (void)buf;
+        // Pokes are allowed, but have no side effects
```
colby-nyce, line 48: (void)size

Obviously this is just a nit, but I'm thinking that plausibly we can/will/should increase the warnings-as-errors and this would start failing to compile.

system/MagicMemory.hpp
```diff
@@ -0,0 +1,50 @@
...
+        //! \brief Name of this resource. Required by sparta::UnitFactory
+        static constexpr char name[] = "MagicMemory";
+
+        class MagicMemoryParameters : public sparta::ParameterSet
+        {
+          public:
+            explicit MagicMemoryParameters(sparta::TreeNode* node) : sparta::ParameterSet(node) {}
+
+            HIDDEN_PARAMETER(uint64_t, base_addr, 0x0, "Base address")
+            HIDDEN_PARAMETER(uint64_t, size, 0, "Memory size")
```
colby-nyce, line 26: Same

system/SimpleUART.hpp
```diff
@@ -0,0 +1,47 @@
...
+      public:
+        //! \brief Name of this resource. Required by sparta::UnitFactory
+        static constexpr char name[] = "SimpleUART";
+
+        class SimpleUARTParameters : public sparta::ParameterSet
+        {
+          public:
+            explicit SimpleUARTParameters(sparta::TreeNode* node) : sparta::ParameterSet(node) {}
+
+            PARAMETER(uint64_t, base_addr, 0x20000000, "Base address")
```
colby-nyce, line 22: Same

colby-nyce (approved):

system/AtlasSystem.cpp
```diff
@@ -77,50 +146,126 @@ namespace atlas
...
+            allocated_blocks.emplace(uart_->getBaseAddr(), uart_->getSize());
+        }
+
+        ////////////////////////////////////////////////////////////////////////////////
+        // Now fill in the memory "blanks"
+        sparta::memory::addr_t addr_block_start = 0;
+        const uint32_t illop = 0;
+        uint32_t block_num = 1;
+
+        while (false == allocated_blocks.empty())
```
lzhu-mips, line 228: For this logic, do we have any safeguard for overlapping blocks? Or we presume all blocks are legit (non-overlapping)?

system/AtlasSystem.cpp
lzhu-mips, line 101: What if the add_delt is 1byte less than ATLAS_SYSTEM_BLOCK_SIZE? Then we would over allocate by ATLAS_SYSTEM_BLOCK_SIZ. I guess we can be a little wasteful here.

system/AtlasSystem.cpp
kathlenemagnus: Didn't know that! Will clean this up.

system/AtlasSystem.cpp
kathlenemagnus: It is not verbose for simple workloads:
```
map: [0         , 0x80001000        ) -> "mb_1        " +0x0
map: [0x80001000, 0x80002000        ) -> "Magic Memory" +0x0
map: [0x80002000, 0x8000000000000000) -> "mb_2        " +0x0
```
Even for more complex workloads, most ELFs won't have more than like 10-15 sections that need mapping.

system/MagicMemory.cpp
kathlenemagnus: I cleaned this issue up in this file and in SimpleUART.cpp.

system/MagicMemory.hpp
```diff
@@ -0,0 +1,50 @@
...
+      public:
+        //! \brief Name of this resource. Required by sparta::UnitFactory
+        static constexpr char name[] = "MagicMemory";
+
+        class MagicMemoryParameters : public sparta::ParameterSet
+        {
+          public:
+            explicit MagicMemoryParameters(sparta::TreeNode* node) : sparta::ParameterSet(node) {}
+
+            HIDDEN_PARAMETER(uint64_t, base_addr, 0x0, "Base address")
```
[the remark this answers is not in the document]
kathlenemagnus: Not a nit at all! Good suggestion.

system/AtlasSystem.cpp
kathlenemagnus: Yah we will have to be a little wasteful in that case since Sparta allocates memory 1 block at a time. We do have control over the block size, but I believe 4K is the smallest size allowed.

kathlenemagnus: I reworked the logic of this section and removed the hardcoded number.

system/AtlasSystem.cpp
kathlenemagnus: ELFs often contain multiple symbols at the same address. For example, the section name `.text` will be in the symbol table with the same address as the first label in that section, often `main`. From what I've seen, the section symbols are listed first followed by the labels and we'd rather use the label name if it's available.

I'm not sure if there's a smart way to do this. Even if I ignore the section symbols, it's still possible that there are multiple labels at the same address.

kathlenemagnus: I added a more detailed comment and added asserts to make sure we don't find multiple tohost/fromhost symbols in an ELF.

system/AtlasSystem.cpp
kathlenemagnus: There are not any safeguards right now. I will add that.

kathlenemagnus: I tested this out and while testing I found out that Sparta ensures that blocks do not overlap. This check is done when `addMapping` is called. I added a comment.

PR #78, opened 2025-02-27
F D extension

lzhu-mips:


core/inst_handlers/d/RvdInsts.hpp
```diff
@@ -3,6 +3,8 @@
 #include <map>
 #include <string>
 
+#include "core/Execute.hpp"
```
lzhu-mips, line 6: I am not sure why the diff shows it here but this is not present on my fork branch.

lzhu-mips: github is showing the diff of the first commit and base. the 2nd commit doesn't have this line.

lzhu-mips:
I'd also like to point out, instead of Post Execute Actions with CSR mapping, we might want to have vector of Actions, unless action for specific CSR can be defined in one place (currently in zicsr handers).

lzhu-pub:
Here are two test suite that failed. fclass is not implemented so fclass test is failing.
For ud: move and ldst failed due to some undefined behavior:
move:  RISCV manual doesn't define fsgnj.s behavior when handling NaN, the test is expecting canonical NaN for some reason.
ldst: RISCV manual doesn't define the upper bit when RV64 flw, the test is expecting all 1's for some reason.

core/inst_handlers/d/RvdInsts.hpp
```diff
@@ -18,6 +20,7 @@ namespace atlas
         static void getInstComputeAddressHandlers(std::map<std::string, Action> & inst_handlers);
         template <typename XLEN>
         static void getInstHandlers(std::map<std::string, Action> & inst_handlers);
+        template <typename XLEN> static void getCsrUpdateActions(Execute::CsrUpdateActionsMap &);
```
colby-nyce, line 23: Split this onto two lines like the above two methods. Better yet: I think you need to run clang-format on your changes.

`find . -name '*.*pp' | xargs clang-format -i`

core/inst_handlers/d/RvdInsts.cpp
```diff
@@ -164,406 +143,197 @@ namespace atlas
...
-        // require_either_extension('D', EXT_ZDINX);
-        // require_fp;
-        // softfloat_roundingMode = RM;
-        // WRITE_FRD_D(i32_to_f64((int32_t)RS1));
-        // set_fp_exceptions;
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        const uint32_t rs1_val = READ_INT_REG<RV64>(state, inst->getRs1());
```
colby-nyce, line 147: It's not clear to me why some of these READ calls are different:

```
const uint32_t val = READ_INT_VAL<RV64>...
const uint64_t val = READ_INT_VAL<RV64>...
```

Is uint32_t only for fcvt_d_w_64_handler?

core/inst_handlers/d/RvdInsts.cpp
```diff
@@ -164,406 +143,197 @@ namespace atlas
...
-        // require_extension('F');
-        // require_fp;
-        // WRITE_FRD(f32(MMU.load<uint32_t>(RS1 + insn.i_imm())));
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        const uint64_t rs1_val = READ_FP_REG<RV64>(state, inst->getRs1());
+        const uint64_t rs2_val = READ_FP_REG<RV64>(state, inst->getRs2());
+        const uint64_t sign_mask = 1UL << 63;
```
colby-nyce, line 294: static constexpr

core/inst_handlers/d/RvdInsts.cpp
```diff
@@ -164,406 +143,197 @@ namespace atlas
...
-
-        // require_extension('D');
-        // require_rv64;
-        // require_fp;
-        // WRITE_FRD(f64(RS1));
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        const uint64_t rs1_val = READ_INT_REG<RV64>(state, inst->getRs1());
```
colby-nyce, line 303: I would really suggest trying "const auto rs1_val" for all these and see if all the passing tests are still passing. The return type of these macros is decided by <RV64> and we should rely on it with "auto".

core/inst_handlers/d/RvdInsts.cpp
```diff
@@ -164,406 +143,197 @@ namespace atlas
...
-        // require_either_extension('D', EXT_ZDINX);
-        // require_fp;
-        // WRITE_FRD_D(fsgnj64(freg(FRS1_D), freg(FRS2_D), true, false));
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        const uint64_t rs1_val = READ_FP_REG<RV64>(state, inst->getRs1());
+        const uint64_t rs2_val = READ_FP_REG<RV64>(state, inst->getRs2());
+        const uint64_t sign_mask = 1UL << 63;
```
colby-nyce, line 334: static constexpr

core/inst_handlers/d/RvdInsts.cpp
```diff
@@ -585,243 +355,133 @@ namespace atlas
...
-        // require_either_extension('D', EXT_ZDINX);
-        // require_fp;
-        // WRITE_FRD_D(fsgnj64(freg(FRS1_D), freg(FRS2_D), false, false));
-
-        // END OF SPIKE CODE
-        ///////////////////////////////////////////////////////////////////////
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        const uint64_t rs1_val = READ_FP_REG<RV64>(state, inst->getRs1());
+        const uint64_t rs2_val = READ_FP_REG<RV64>(state, inst->getRs2());
+        const uint64_t sign_mask = 1UL << 63;
```
colby-nyce, line 445: static constexpr

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,143 @@
...
+{
+#include "source/include/softfloat.h"
+}
+
+namespace atlas
+{
+    using SP = uint32_t;
+    using DP = uint64_t;
+
+    template <typename XLEN> inline uint_fast8_t getRM(const AtlasInstPtr & inst, AtlasState* state)
```
colby-nyce, line 19: You could clean up the call sites with just `getRM(AtlasState* state)` since you can get the AtlasInstPtr from it.

core/inst_handlers/d/RvdInsts.cpp
lzhu-pub: Whether it's uint32_t or uint64_t depending on the  instruction. Here fcvt_d_w convert from word 32 bit to double precision. So it's stored in uint32_t.

core/inst_handlers/d/RvdInsts.cpp
lzhu-pub: please see previous comment.

core/inst_handlers/d/RvdInsts.hpp
lzhu-pub: I run it with CMkae ---target clangformat. it doesn't change the above line.
Maybe it's not running the same clang-format.

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    template <typename XLEN> static ActionGroup* update_csr(AtlasState* state)
+    {
+        // TODO: it would be nice to have field shift, then a single combined CSR write will
+        // suffice.
+
+        // FFLAGS
+        WRITE_CSR_FIELD<XLEN>(
```
colby-nyce, line 74: I can't decide what I want to do with this code.... but this is taking an unnecessary performance hit with all these WRITE_CSR_FIELD calls. Especially since update_csr() is called every time. 

Maybe forget about performance for now, but we should take a note to optimize this further.

core/inst_handlers/f/inst_helpers.hpp
```diff
@@ -0,0 +1,143 @@
...
+            state, FCSR, "DZ",
+            static_cast<uint64_t>((softfloat_exceptionFlags & softfloat_flag_infinite) != 0));
+        WRITE_CSR_FIELD<XLEN>(
+            state, FCSR, "NV",
+            static_cast<uint64_t>((softfloat_exceptionFlags & softfloat_flag_invalid) != 0));
+
+        return nullptr;
+    }
+
+    template <typename T> ActionGroup* compute_address_handler(AtlasState* state)
```
colby-nyce, line 110: Keep all the template/typename the same with XLEN.

core/inst_handlers/zicsr/RvzicsrInsts.cpp
```diff
@@ -274,4 +301,73 @@ namespace atlas
 
         return nullptr;
     }
+
+    template <typename XLEN> void set_softfloat_excpetionFlags(atlas::AtlasState* state)
+    {
+        XLEN softflaot_exceptionFlags_mask = softfloat_flag_inexact | softfloat_flag_underflow
```
colby-nyce, line 307: typo (flaot)

colby-nyce (approved):

colby-nyce:
Please wait for Kathlene's feedback before merging.

core/inst_handlers/d/RvdInsts.hpp
lzhu-pub: the command you provided didn't change the line either. I will fix it.

core/inst_handlers/f/inst_helpers.hpp
kathlenemagnus, line 19: I would prefer all of these functions to be private methods of `RvfInsts` assuming they're only called by the F instructions.

core/inst_handlers/d/RvdInsts.cpp
```diff
@@ -164,406 +143,197 @@ namespace atlas
...
+        update_csr<RV64>(state);
         return nullptr;
     }
 
     ActionGroup* RvdInsts::fld_64_compute_address_handler(atlas::AtlasState* state)
     {
-        state->getCurrentInst()->markUnimplemented();
-        (void)state;
-        return nullptr;
+        return compute_address_handler<RV64>(state);
```
kathlenemagnus, line 281: For these instruction handlers where you are only calling another function, you can just use the called function (`compute_address_handler`) as the instruction handler directly. There's no reason to wrap the function call.
```
inst_handlers.emplace(
                "fld",
                atlas::Action::createAction<&RvdInsts::compute_address_handler<RV64>, RvdInsts>(
                    nullptr, "fld", ActionTags::COMPUTE_ADDR_TAG));
```

lzhu-pub: It is called by F and D both. So it's not a member function yet. I will see what I can do.

core/inst_handlers/f/inst_helpers.hpp
lzhu-pub: they are called by F and D.

PR #81, opened 2025-03-17
More Work to Support RV32

kathlenemagnus-mips:
- Refactored the RVA instruction handler class to reduce code duplication (deleted `inst_helpers.hpp`)
- Defined RV32 RVA instruction handlers
- Defined RV32 translation Actions
- Extended the Instruction Logger to print RV32 values correctly
- Fixed several branch handlers that were using the wrong types for RV32
- Updated the ecall instruction handler to support RV32


kathlenemagnus-mips:
With these fixes we now have some RV32 tests passing! Currently, 57/99 arch tests are passing.

lzhu-pub (approved):
I am good with the template for RV32 and RV64.

The only comment I'd like to point out is, now we are doing something like:
if constexpr (RV64):
    inst_handlers.empalce(func\<RV64\>, "inst")
else: (RV32)
    inst_handlers.empalce(func\<RV32\>, "inst")
    
The instruction string here are the same, but just for different RVs.
Is it  possible to directly forward "XLEN" from  template <XLEN> void RvXXX::getInstHandlers(std::map<std::string, Action> & inst_handlers) to the templated functions?

Some instructions may only support RV64 or RV32, but we can do an "if" check  just for those instructions.

kathlenemagnus:
> I am good with the template for RV32 and RV64.
> 
> The only comment I'd like to point out is, now we are doing something like: if constexpr (RV64): inst_handlers.empalce(func<RV64>, "inst") else: (RV32) inst_handlers.empalce(func<RV32>, "inst")
> 
> The instruction string here are the same, but just for different RVs. Is it possible to directly forward "XLEN" from template void RvXXX::getInstHandlers(std::map<std::string, Action> & inst_handlers) to the templated functions?
> 
> Some instructions may only support RV64 or RV32, but we can do an "if" check just for those instructions.

That's a great idea. Let me try it in this PR for RVA.

kathlenemagnus:
Made a few more fixes and now the RV64 arch test pass rate is 118/130.

PR #87, opened 2025-03-21
issue 84 translation request cleanup (part 1 of DMI)

### LINKED ISSUE #84: Simplify instruction translation
Instruction and Data translation occur using data structures captured in the AtlasState object, specifically via
an object called `AtlasTranslationState`.  Both Fetch and Execute (more specifically, instructions) use this singular state
instance to "communicate" the requested VA to be translated either for fetching the next instruction or loading/storing data.

The problem with using the same translation object is this prevents the simulator from taking full advantage of DMI accesses
to those areas of memory that represent the pages translated.  In other words, for fetch, the DMI region can be (and most likely will be) _different_ from the DMI region for executed loads/stores.

Suggested fix:

- AtlasState's current "translation state" is for instruction access only
- Have each instruction maintain its own translation request/response objects

Once these separations are made, the behaviors of fetches/instructions become:

- Fetch can use AtlasState for current translation response for the given address.  If still usable, use DMI to get the next instruction
- Load/store handlers can use their translation responses (after a verification of that response by translation) to get the DMI pointer

klingaard:
TranslationState now resides in two places: AtlasState and AtlasInst.
 
- Fetch requests for translation are made via the state class
- Load/store/atomic/etc, i.e. instruction based translation occurs via the translation state contained in the instruction
- Simplified the API a bit and cleaned up some code
- Moved MMU mode directly into AtlasState 
- Regression: PASS RATE: 118/130

closes #84


CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ project(Atlas CXX)
 set (SIM_BASE ${PROJECT_SOURCE_DIR})
 
 # Profile build flags
-set(CMAKE_CXX_FLAGS_FASTDEBUG "-O3 -g")
+set(CMAKE_CXX_FLAGS_FASTDEBUG "-O3 -g -DNDEBUG")
```
klingaard, line 24: This bit me again.  My latest conda env toggles this value between some of the CMakeLists.  Dunno why.  Force the value.

core/translate/Translate.cpp
```diff
@@ -123,39 +93,51 @@ namespace atlas
...
-        const AtlasTranslationState::TranslationRequest request =
-            translation_state->getTranslationRequest();
-        const XLEN vaddr = request.virtual_addr;
-        const uint32_t width = std::is_same_v<XLEN, RV64> ? 16 : 8;
-        ILOG("Translating " << HEX(vaddr, width));
+        AtlasTranslationState* translation_state = nullptr;
+        if constexpr (INST_TRANSLATION)
+        {
+            // Translation reqest is from fetch
+            translation_state = state->getFetchTranslationState();
```
kathlenemagnus, line 103: If we want to deprecate the `TranslationState` class, why not just call "state->getPc()" instead?

core/translate/Translate.hpp
```diff
@@ -35,9 +35,10 @@ namespace atlas
 
         void changeMmuMode(const uint64_t xlen, MMUMode mode);
 
-      private:
-        template <typename XLEN, MMUMode Mode> ActionGroup* translate_(atlas::AtlasState* state);
+        MMUMode getMode() const { return mode_; }
 
+      private:
+        MMUMode mode_ = MMUMode::BAREMETAL;
```
kathlenemagnus, line 41: Why did you need to add this? I would keep it in `AtlasState` somewhere.

core/translate/Translate.cpp
kathlenemagnus: Just to clarify, I think we should deprecate `TranslationState` for making translation request and returning results. It's still useful for managing the MMU mode and caching some CSR values and should remain a part of `AtlasState`.

kathlenemagnus (approved):
This is great. I see the vision! 👍

core/translate/Translate.hpp
klingaard: I actually moved it from TranslationState to Translate in case something, somewhere needed this info.  But I agree, it can be in AtlasState.

core/translate/Translate.cpp
klingaard: Yeah, I want to remove it as well, but I ran into a small snag for fetch: where do I store the results of the translation?  I either stash just the translation results in the Atlas state or use the existing `TranslationState` class and not change any of the code below this `constexpr`  For now, I; choosing the lazy path.  

"Part 2" of this effort I will look at that.

core/translate/Translate.hpp
klingaard: Done

PR #91, opened 2025-03-24
Supervisor Exceptions

kathlenemagnus:
Added support for taking exceptions in Supervisor mode. This is controlled by the MEDELEG/SEDELEG CSRs for exceptions and the MIDELEG/SIDELEG CSRs for interrupts. I also did a little bit of ground work needed to support interrupts in the future. With this PR, the "sbreak" arch test is now working correctly.


core/Exception.cpp
```diff
@@ -30,77 +30,119 @@ namespace atlas
...
+        switch (cause)
+        {
+            case InterruptCause::SUPERVISOR_SOFTWARE:
+            case InterruptCause::MACHINE_SOFTWARE:
+            case InterruptCause::SUPERVISOR_TIMER:
+            case InterruptCause::MACHINE_TIMER:
+            case InterruptCause::SUPERVISOR_EXTERNAL:
+            case InterruptCause::MACHINE_EXTERNAL:
+            case InterruptCause::COUNTER_OVERFLOW:
+            default:
```
knute-mips, line 143: Remove `default`.  Not needed.

core/Exception.cpp
```diff
@@ -30,77 +30,119 @@ namespace atlas
...
-            case TrapCauses::MACHINE_ECALL:
-            case TrapCauses::SOFTWARE_CHECK_FAULT:
-            case TrapCauses::HARDWARE_ERROR_FAULT:
+            case FaultCause::BREAKPOINT:
+            case FaultCause::USER_ECALL:
+            case FaultCause::SUPERVISOR_ECALL:
+            case FaultCause::MACHINE_ECALL:
+            case FaultCause::SOFTWARE_CHECK:
+            case FaultCause::HARDWARE_ERROR:
             default:
```
knute-mips, line 126: Remove `default`

core/Exception.cpp
```diff
@@ -30,77 +30,119 @@ namespace atlas
...
-        const uint64_t cause_val = static_cast<uint64_t>(cause_.getValue());
-        WRITE_CSR_REG<XLEN>(state, MCAUSE, cause_val);
+        const uint32_t cause_csr = (priv_mode == PrivMode::SUPERVISOR) ? SCAUSE : MCAUSE;
+        WRITE_CSR_REG<XLEN>(state, cause_csr, cause_val);
 
         // Depending on the exception type, get the
-        const uint64_t trap_val = determineTrapValue_(cause_.getValue(), state);
-        WRITE_CSR_REG<XLEN>(state, MTVAL, trap_val);
+        const uint32_t tval_csr = (priv_mode == PrivMode::SUPERVISOR) ? STVAL : MTVAL;
+        const uint64_t trap_val = is_interrupt ? determineTrapValue_(interrupt_cause_.getValue(), state) : determineTrapValue_(fault_cause_.getValue(), state);
```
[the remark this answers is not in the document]
kathlenemagnus: Yahhhh I was playing with the best way to reduce code duplication. I would prefer just 1 priv mode check, too. I'll make that change.

core/Exception.cpp
```diff
@@ -30,77 +30,139 @@ namespace atlas
 
     template <typename XLEN> ActionGroup* Exception::handleException_(atlas::AtlasState* state)
     {
-        sparta_assert(cause_.isValid(), "Exception cause is not valid!");
+        sparta_assert(fault_cause_.isValid() || interrupt_cause_.isValid(),
```
colby-nyce, line 33: Is there any sense in assert(fault_cause_.isValid() != interrupt_cause_.isValid()) as well?

core/Exception.cpp
```diff
@@ -30,77 +30,139 @@ namespace atlas
...
+        switch (cause)
         {
-            case TrapCauses::MISALIGNED_FETCH:
-            case TrapCauses::FETCH_ACCESS:
-            case TrapCauses::FETCH_PAGE_FAULT:
+            case FaultCause::INST_ADDR_MISALIGNED:
+            case FaultCause::INST_ACCESS:
+            case FaultCause::INST_PAGE_FAULT:
                 return state->getPc();
                 break;
```
colby-nyce, line 130: break statements are not needed when preceded by return statement

core/Exception.cpp
kathlenemagnus-mips: Yah that would be good to add.

PR #92, opened 2025-03-25
Add system memory writes/reads to the inst logger

colby-nyce:


core/observers/Observer.hpp
```diff
@@ -76,6 +111,56 @@ namespace atlas
...
+            }
+
+            MemRead mem_read;
+            mem_read.addr = data.addr;
+            mem_read.size = data.size;
+            mem_read.value = val;
+            mem_read_ = mem_read;
+        }
+
+        sparta::utils::ValidValue<MemRead> mem_read_;
```
kathlenemagnus, line 162: An instruction can cause more than 1 read or write to memory, so these needs to be vectors. For example, a misaligned load will cause 2 reads to memory. Load/store pair and vector instructions can also trigger multiple reads and writes.

core/AtlasState.cpp
```diff
@@ -194,6 +194,22 @@ namespace atlas
...
+        {
+            obs->registerFor(atlas_system_->getSystemMemory());
+        }
+    }
+
+    ActionGroup* AtlasState::stopSim_(AtlasState*)
+    {
+        for (auto & obs : observers_)
+        {
+            obs->deregisterFor(atlas_system_->getSystemMemory());
```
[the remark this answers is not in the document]
colby-nyce: Removed

core/observers/Observer.hpp
colby-nyce: Fixed

kathlenemagnus (approved):

core/observers/Observer.hpp
```diff
@@ -75,7 +96,54 @@ namespace atlas
...
+            if (data.prior)
+            {
+                for (size_t i = 0; i < data.size; ++i)
+                {
+                    prior_val |= static_cast<uint64_t>(data.prior[i]) << (i * 8);
+                }
+            }
+
+            uint8_t buf[2048];
+            data.mem->peek(data.addr, data.size, buf);
```
knute-mips, line 117: Shouldn't you have an assert that `data.size` < 2048?

PR #94, opened 2025-03-27
Expose memory reads/writes to the IDE

colby-nyce:
There is room for improvement here on how we show the memory reads/writes that occur during an instruction handler, but this first submission will help close out the remaining ISA tests. The reads and writes are just listed along the top "banner" with the RD/CSR changes for now. They are not given their own grid/table like the others since we haven't really discussed if that's appropriate or just clutter.

### DIFF
--- a/core/observers/SimController.cpp
+++ b/core/observers/SimController.cpp
@@ -295,6 +295,9 @@ namespace atlas
                                const std::vector<std::string> & args,
                                ActionGroup*& fail_action_group)
         {
+            // TODO cnyce: Consider replacing all use of stringstream here to
+            // only send hex values as uint64_t and format in Python.
+
             switch (sim_cmd)
             {
                 case SimCommand::XLEN:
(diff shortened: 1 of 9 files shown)


core/observers/SimController.cpp
```diff
@@ -646,6 +654,57 @@ namespace atlas
...
+                        return true;
+                    }
+
+                case SimCommand::MEM_WRITES:
+                    {
+                        std::string json;
+                        for (size_t idx = 0; idx < mem_writes_.size(); ++idx)
+                        {
+                            const auto & mem_write = mem_writes_[idx];
+                            std::stringstream ss;
```
knute-mips, line 688: If you start running into performance issues, one thing to note is that `stringstream` is ridiculously slow.  Consider sending the data over as int and doing the hex conversions on the rendering side.

knute-mips (approved):

klingaard (approved):

PR #95, opened 2025-03-31
Supports Writes to MISA + Better CSR Instruction Handling

kathlenemagnus:
Started out by implementing the MISA CSR's update handler and found a bunch of other things to fix along the way:
- Writes to the MISA CSR now change which extensions are enabled with Mavis
- Write to the FS field of the MSTATUS CSR now enables/disables the floating point extensions
- Only insert CSR update Actions for instructions that actually update the CSR register
- Enabled the C extension for compressed instruction support
- Check if reads and/or writes to a CSR register are allowed, throw an illegal inst exception if not


core/AtlasInst.cpp
```diff
@@ -67,6 +67,37 @@ namespace atlas
     {
     }
 
+    bool AtlasInst::writesCsr() const
+    {
+        if (SPARTA_EXPECT_FALSE(hasCsr()))
```
knute-mips, line 72: Curious, can all of this inquiry be done in the extracted inst?

core/AtlasState.cpp
```diff
@@ -201,6 +178,21 @@ namespace atlas
         }
     }
 
+    mavis::FileNameListType AtlasState::getUArchFiles_()
```
knute-mips, line 181: `const`?  🍺

core/AtlasState.cpp
```diff
@@ -432,6 +424,24 @@ namespace atlas
         return 0;
     }
 
+    template <typename XLEN> uint32_t AtlasState::getMisaExtFieldValue_()
```
knute-mips, line 427: `const`?  🍻

knute-mips (approved):

core/AtlasInst.cpp
kathlenemagnus: Some of it could be, but we still have to look at the rs1 number and the immediate value. I'd rather keep the logic all in one function.

core/AtlasInst.cpp
```diff
@@ -67,6 +67,37 @@ namespace atlas
...
+        {
+            const auto mavis_uid = getMavisUid();
+            // csrrw/csrrwi always writes
+            // csrrs/csrrc write if rs1 != 0
+            // csrrsi/csrrci writes if imm != 0
+            if ((mavis_uid == AtlasState::MavisUIDs::MAVIS_UID_CSRRW)
```
colby-nyce, line 78: Use a switch for performance

core/AtlasState.cpp
```diff
@@ -439,6 +431,24 @@ namespace atlas
         return 0;
     }
 
+    template <typename XLEN> uint32_t AtlasState::getMisaExtFieldValue_() const
+    {
+        uint32_t ext_val = 0;
+        for (auto & ext : extension_manager_.getEnabledExtensions())
+        {
+            if (ext.first.size() == 1)
```
colby-nyce, line 439: I think this deserves a comment why size()==1 is special (either that or a different API added alongside getEnabledExtensions)

core/AtlasState.cpp
```diff
@@ -439,6 +431,24 @@ namespace atlas
...
+            {
+                ext_val |= 1 << getCsrBitRange<XLEN>(MISA, &ext.first[0]).first;
```
colby-nyce, line 441: Can you use `ext.first.c_str()` ?

core/inst_handlers/zicsr/RvzicsrInsts.cpp
```diff
@@ -134,10 +99,19 @@ namespace atlas
         auto rd = insn->getRd();
         const uint32_t csr = insn->getCsr();
 
+        if (!isAccessLegal_(csr, state->getPrivMode()))
```
colby-nyce, line 102: I'm confused how this compiles since the template arg WRITE does not default to false (or rather doesn't default to anything). I would prefer to use `enum AccessType { WRITE, READ };` than a boolean as well.

core/inst_handlers/zicsr/RvzicsrInsts.cpp
```diff
@@ -249,13 +269,64 @@ namespace atlas
         return nullptr;
     }
 
+    template <typename XLEN>
+    ActionGroup* RvzicsrInsts::misa_update_handler(atlas::AtlasState* state)
+    {
+        const XLEN misa_val = READ_CSR_REG<XLEN>(state, MISA);
+        auto & ext_manager = state->getExtensionManager();
```
colby-nyce, line 276: It looks like only `RISCVExtensionManager::isEnabled()` is used, so getExtensionManager() should be a const method returning const.

colby-nyce (approved):

scripts/RV64_CSR.py
```diff
@@ -257,29 +257,29 @@
         "WLRL" : { 'high_bit' : 61, 'low_bit' : 26, 'readonly' : False, "desc" : "Unknown" },
         "Extensions" : { 'high_bit' : 25, 'low_bit' : 0,  'readonly' : False, "desc" : "Encodes the native base integer ISA" },
         "A" : { 'high_bit' : 0 , 'low_bit' : 0 ,  'readonly' : False, "desc" : "Atomic extension" },
-        "B" : { 'high_bit' : 1 , 'low_bit' : 1 ,  'readonly' : True,  "desc" : "Bit Manipulation extension" },
-        "C" : { 'high_bit' : 2 , 'low_bit' : 2 ,  'readonly' : False, "desc" : "Compressed extension" },
+        "B" : { 'high_bit' : 1 , 'low_bit' : 1 ,  'readonly' : False, "desc" : "Bit Manipulation extension" },
+        "C" : { 'high_bit' : 2 , 'low_bit' : 2 ,  'readonly' : False, "dtesc" : "Compressed extension" },
```
lzhu-pub, line 261: typo?

kathlenemagnus: Fixed! Good catch.

Issue #97, opened 2025-04-01
Add support for changing address translation modes


skhanx86:
While executing the test [icache-alias.S](https://github.com/riscv-software-src/riscv-tests/blob/master/isa/rv64si/icache-alias.S) on the ATLAS simulator, it was observed that the test changes the address translation mode during its execution. However, ATLAS currently only supports BareMetal (va=pa) translation mode by default.

ATLAS has the page table translation logic in place but lacks the ability to dynamically switch between different address translation modes.

Expected Behavior:
The simulator should allow changing the address translation modes (e.g., BareMetal, Sv32, Sv39, etc.) during execution as required by the test or application.

Current Behavior:
The simulator only supports the BareMetal mode (va=pa) and does not provide the ability to switch between address translation modes dynamically.

Additional Context:
The issue was identified while running the aforementioned RISC-V test, which involves address translation.

knute-mips:
I have a branch where I'm fixing a few issues with translation and attempted to run this test.  In my branch, Atlas can now switch between the different translation modes and gets much farther in this test.  However, Atlas does not discern between and instruction translation fault and a load/store page fault yet.  This will be the next task after I merge in my branch.

Issue #12, opened 2024-11-18
Add Support for Sv32 Translation


kathlenemagnus-mips:

klingaard:
To make this work efficiently, we really need support for #11 .  As an example, when `satp` is written, the Translate unit should be notified and positions itself to handle the proper walking levels, PTE info, etc once the machine mode changes.

PR #98, opened 2025-04-02
Translation fixes

klingaard:
- Got translation support _mostly_ working, but tests are still not passing; I need to pause for now
- Added basic framework for hypervisor support (but ignored)
- Added medeleg fields

Addresses issues #97 #12


system/MagicMemory.cpp
```diff
@@ -67,7 +67,7 @@ namespace atlas
                 }
                 break;
             case SupportedDevices::BLOCK_CHAR:
-                sparta_assert(false, "TODO: Block char is unsupported");
+                //sparta_assert(false, "TODO: Block char is unsupported");
```
klingaard, line 70: This assert is firing with the -v- tests; I commented it out for now to see how far the tests go.  They still all fail. :laughing:

klingaard:
I implemented block char writes and I get this when running one of the -v- tests:
```
MAGICMEM: Assertion failed: pt[3][addr/(1UL << 12)] & 0x040
```
Neat.

system/MagicMemory.cpp
```diff
@@ -67,7 +67,20 @@ namespace atlas
...
+                {
+                    sparta_assert(false, "TODO: Block char read is unsupported");
+                }
+                else
+                {
+                    const char letter = char(mm_command.tohost_data);
+                    block_char_msg_ << letter;
+                    if (letter == '\n')
+                    {
+                        std::cout << "MAGICMEM: " << block_char_msg_.str();
```
klingaard, line 80: @kathlenemagnus would prefer this go to `std::cerr`?

kathlenemagnus: `std::cout` is fine. Do we need to include `MAGICMEM`? Since it only writes 1 character at a time I think the formatting will look weird.

core/AtlasState.cpp
```diff
@@ -47,6 +47,7 @@ namespace atlas
             supported_isa_string_, isa_file_path_ + std::string("/riscv_isa_spec.json"),
             isa_file_path_)),
         stop_sim_on_wfi_(p->stop_sim_on_wfi),
+        hypervisor_enabled_(extension_manager_.isEnabled("h")),
```
kathlenemagnus, line 50: Can you add an assert in the constructor to make sure this is false? I don't want someone to try to turn it on and think it's supported.

core/AtlasState.cpp
```diff
@@ -185,6 +186,11 @@ namespace atlas
         }
     }
 
+    void AtlasState::changeMMUMode(uint32_t satp_mode)
```
kathlenemagnus, line 189: Use `MMUMode` enum defined in `include/AtlasTypes.hpp`.

core/AtlasState.cpp
```diff
@@ -536,10 +542,6 @@ namespace atlas
         pc_ = next_pc_;
         DLOG("PC: 0x" << std::hex << pc_);
 
-        // Set Privilege Mode
-        priv_mode_ = next_priv_mode_;
```
kathlenemagnus, line 540: Did you change this because you needed to change the privilege mode and immediately see the effect? If so, maybe we can add a "getNextPrivMode` method? I haven't done this yet, but I want to print the privilege mode in the inst logger and I need to know the privilege mode that the instruction was executed in.

core/AtlasState.cpp
klingaard: Done

core/AtlasState.cpp
klingaard: This is the raw satp mode which unfortunately does not map to MMUMode types...  

The function `changeMMUMode` in Translate does the mapping

klingaard: I can rearrange that so we don't pass around raw satp values...

system/MagicMemory.cpp
klingaard: Done

core/AtlasState.cpp
klingaard: TBH I didn't understand why the priv mode was cached until the PC was updated.  The places that change the priv mode are exception handling and xret, which at the finish of their execution the new priv mode was now in effect.  

When xret completes for example, the next instruction's priv mode is apparent.  I think the logger would look like:

```
0xFFFFFFFF80000100  priv:MACHINE set 
0x0000000000000100  priv:USER addi
```

core/AtlasState.cpp
kathlenemagnus: Can the SATP mode be mapped to the MMUMode types by changing the value of the enum? Like the way we have the PrivMode enum set up so we can do static casts.

core/translate/Translate.cpp
```diff
@@ -92,11 +126,11 @@ namespace atlas
         }
     }
 
-    template <typename XLEN, MMUMode MODE, bool INST_TRANSLATION>
+    template <typename XLEN, MMUMode MODE, bool TRANSLATION>
```
kathlenemagnus, line 129: You don't need to include this in your PR, but we will need to change this bool to some new enum for the access type (inst, load or store) so we can throw the appropriate exception type.

kathlenemagnus (approved):

core/translate/Translate.cpp
klingaard: Already thought of that.  :grin:   I actually started implementing that so I could remove the `is_store_type_` in AtlasInst.hpp, but decided to go ahead with this PR

core/AtlasState.cpp
klingaard: MMUMode is used as an index into multiple arrays, unfortunately.  We'd either have to change the arrays to maps or make sparse arrays.  

I think a generic conversion utility is in order.  I can do that in the next PR

PR #99, opened 2025-04-03
Clangformat Fix

kathlenemagnus:
Added support to require a minimum clang-format version. Unfortunately, I can't raise the minimum any higher that v15; several developers use WSL and its Ubuntu version only supports up to v15.


lzhu-pub (approved):

klingaard:
Neat.  All green.  Guess we just needed to make a simplified conda environment.   But I think the main issue was that we were using a combination of conda and local compilers -- that's no bueno.  `gcc` was not part of the conda distribution.

klingaard:
The ubuntu image we're using has three clangformats to use: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

```
Clang-format: 16.0.6, 17.0.6, 18.1.3
```
But let's install a specific one, clangformat-14.  We can move if we want to

PR #103, opened 2025-04-04
 Fixed a few translation related issues (mostly permissions)

klingaard:
Added PTE permission checking and cleaned up the code a bit.

New pass rate with --no-skip-virt flag to RunRiscVArchTest.py:

PASS RATE: 199/235

### DIFF
--- a/core/translate/PageTableEntry.hpp
+++ b/core/translate/PageTableEntry.hpp
@@ -68,7 +68,17 @@ namespace atlas
         // Valid bit, same fpr all modes
         bool isValid() const { return pte_val_ & translate_types::Sv32::PteFields::valid.bitmask; }
 
+        // Is this a Leaf PTE entry
         bool isLeaf() const { return canRead() || canExecute(); }
+
+        // Is this PTE accessable to the type
+        bool isAccessable(bool store) const
+        {
+            const decltype(pte_val_) access_mask =
+                translate_types::Sv32::PteFields::accessed.bitmask
+                | (store ? translate_types::Sv32::PteFields::dirty.bitmask : 0);
+            return pte_val_ & access_mask;
+        }
 
       private:
         XLEN pte_val_;
--- a/core/translate/PageTableEntry.hpp
+++ b/core/translate/PageTableEntry.hpp
@@ -29,11 +29,17 @@ namespace atlas
         // Dirty bit, same for all modes
         bool isDirty() const { return pte_val_ & translate_types::Sv32::PteFields::dirty.bitmask; }
 
+        // Set the page dirty
+        void setDirty() { pte_val_ |= translate_types::Sv32::PteFields::dirty.bitmask; }
+
         // Accessed bit, same for all modes
         bool isAccessed() const
         {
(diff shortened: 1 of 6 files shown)


core/translate/PageTableEntry.hpp
```diff
@@ -62,8 +68,18 @@ namespace atlas
         // Valid bit, same fpr all modes
         bool isValid() const { return pte_val_ & translate_types::Sv32::PteFields::valid.bitmask; }
 
+        // Is this a Leaf PTE entry
         bool isLeaf() const { return canRead() || canExecute(); }
 
+        // Is this PTE accessable to the type
+        bool isAccessable(bool store) const
```
lzhu-pub, line 75: I don't quit get where this logic is coming from. Is this documented in RISCV manual?

core/translate/Translate.cpp
```diff
@@ -179,25 +181,23 @@ namespace atlas
...
-                    THROW_LOAD_PAGE_FAULT;
                 }
+
+                break;
             }
 
             // If PTE is a leaf, perform address translation
             if (pte.isLeaf())
             {
+                if ((false == pte.isUserMode()) && (priv_mode != PrivMode::SUPERVISOR))
```
lzhu-pub, line 195: Do you mean `PrivMode::USER`?

core/translate/Translate.cpp
```diff
@@ -211,6 +211,24 @@ namespace atlas
...
+                        pte.setAccessed();
+                        state->writeMemory<XLEN>(pte_paddr, pte.getPte());
+                    }
+                    else
+                    {
+                        // Take exception -- no access allowed or not dirty
+                        break;
+                    }
+                }
                 const auto index_bits = (vpn_field.msb - vpn_field.lsb + 1) * indexed_level;
```
lzhu-pub, line 232: I don't quite understand the line here.

core/translate/PageTableEntry.hpp
knute-mips: I think this is an accessibility statement:
> Section 12.3.1: The Svade extension is enabled: when a virtual page is accessed and the A bit is clear, or is written and the D bit is clear, a page-fault exception is raised.
> 
> When the Svade extension is not implemented: update the A/D bits based on rules...

I think the RV compliance tests were expecting `svade` extension to be implemented.

core/translate/Translate.cpp
knute-mips: I could be a little smarter here and grab those VPN fields from the previous VPN index before hitting the leaf (i.e. build the virtual base on the fly).  But this was a cute trick I saw in Spike.  Basically, build a 9 bit mask, shift it based on where we are in the look up, and "or" in those bits from the vaddr.

core/translate/Translate.cpp
```diff
@@ -244,6 +262,7 @@ namespace atlas
         {
             THROW_LOAD_PAGE_FAULT;
         }
+        return nullptr;
```
kathlenemagnus, line 265: Can you add a comment up on line 251 that explains that if we break out of the loop above then an exception needs to be thrown.

PR #104, opened 2025-04-08
InstructionLogger formatted similar to Spike

colby-nyce:


IDE/backend/sim_wrapper.py
```diff
@@ -22,7 +22,14 @@ def UnscopedEnter(self):
         riscv_tests_dir = os.path.abspath(self.riscv_tests_dir)
         os.chdir(os.path.dirname(self.sim_exe_path))
         program_path = "./atlas"
-        program_args = ["--interactive", f"{riscv_tests_dir}/{self.test_name}"]
+
+        program_args = ["--interactive"]
+        if self.test_name.startswith('rv32'):
+            program_args.extend(["-p", "top.core0.params.isa_string", "rv32g_zicsr_zifencei"])
+        else:
+            program_args.extend(["-p", "top.core0.params.isa_string", "rv64g_zicsr_zifencei"])
```
knute-mips, line 30:
```suggestion
        arch = self.test_name[0:3]
        assert arch == 'rv32' or arch == 'rv64'
        program_args.extend(["-p", "top.core0.params.isa_string", f"{arch}g_zicsr_zifencei"])
```

core/observers/InstructionLogger.cpp
knute-mips, line 1: Hmmmm... not sure I like this.  There's a LOT of duplicate code between a spike logger and the generate Atlas logger.  

Why didn't you just make a custom sparta::MessageSource and override the stringize function?  That way none of this code really changes...

knute-mips (changes requested):

core/observers/InstructionLogger.cpp
colby-nyce: Now that I've gone deeper into other designs / uses of MessageSource, I'm not really seeing what you mean by using the emit_ function and I don't see how the InstructionLogger code cannot change (atlasPostExecute vs spikePostExecute).

I can certainly reduce duplicate code between those two functions, but how can we avoid using two separate functions completely? For instance, Atlas might say:

INSTLOG("imm: " << HEX16(imm_val));

But Spike doesn't print out immediate values, so right away I object that the existing calling code cannot really change. On top of that, nothing in MessageSource/NotificationSource seems to be the function that you want to override. The emit_ function (if I made it virtual) is already taking a string, and it is the _string itself_ that I want to "override", not _where the string is written_, which is what overriding emit_ would do.

I can certainly go back to using MessageSource, but I'm not seeing how I can customize / subclass MessageSource at all. I think it still boils down to refactoring to eliminate duplicate code and to remove the MessageLogger.hpp I added. Am I missing something here?

knute-mips: I guess my thinking is to use `sparta::log::MessageSource` as a _template_ for a new logger: `SpikeLogger` which derives from `NotificationSource<sparta::log::Message>`.  You collect the string (using `operator<<`) for the logger and post the string as a notification to emit.

colby-nyce: I'm going to close out this PR since the refactoring that I've done is better than this. But I'll look into the SpikeLogger suggestion if the next PR that I open still needs some refinement.

PR #105, opened 2025-04-08
Attempt a simplified build environment

klingaard:
Removed the need for conda.  The Ubuntu CI machines (runner images) mostly have everything we need, so let's just use them.


colby-nyce:
Would `conda activate sparta` still be needed if we are building Atlas and Sparta together, such as to get a single debug build for everything? (i.e. build Sparta in debug mode but do not install it anywhere).

klingaard:
> Would `conda activate sparta` still be needed if we are building Atlas and Sparta together, such as to get a single debug build for everything? (i.e. build Sparta in debug mode but do not install it anywhere).

Not really.  `conda` is handy for really one thing: we're stuck on machines that have software installed from the 90s.  You don't need conda on your local machine allowing you to do whatever you want.  If you want to install sparta in `$HOME/my_sparta_debug` and `$HOME/my_sparta_reelease` go for it.

The risk we might be running is you (Colby :smile: ) might require mysqlite v4 (making stuff up), but the rest of us are using v3.  You can "fix" that by using `cmake` rules that say, MUST USE v4!!!  And cmake will barf.

kathlenemagnus (approved):

PR #110, opened 2025-04-08
Fix for Failing RV32 Tests

kathlenemagnus:
Several miscellaneous fixes:
- Fixed failing RV32 tests by accessing some mstatus fields through mstatush (RV32 only CSR)
- Fixed build issues when python scripts for register generation are changed
- Catch all Mavis exceptions when decoding fails
- Updated python script for running arch tests with new parameters (`--rv32-only`, `--rv64-only`, `--p-only`, `--v-only`)


core/Fetch.cpp
```diff
@@ -89,7 +89,7 @@ namespace atlas
             // Set next PC, can be overidden by a branch/jump instruction or an exception
             state->setNextPc(state->getPc() + opcode_size);
         }
-        catch (const mavis::UnknownOpcode & e)
+        catch (const mavis::BaseException & e)
```
knute-mips, line 92: Curious, why did you do this?

klingaard (approved):

PR #119, opened 2025-04-14
v ext: add vset{i}vl{i} instructions

lzhu-mips:
added vset{i}vl{i} inst_handlers.
added unit tests for above handlers.
changed READ/WRITE_VEC_REG macro.

### DIFF
--- a/core/AtlasInst.hpp
+++ b/core/AtlasInst.hpp
@@ -115,6 +115,11 @@ namespace atlas
             return 0;
         }
 
+        uint64_t getAVL() const
+        {
+            return opcode_info_->getSpecialField(mavis::OpcodeInfo::SpecialField::AVL);
+        }
+
         sparta::Register* getRs1Reg() const
         {
             sparta_assert(rs1_reg_, "Operand RS1 is a nullptr! " << *this);
--- a/core/AtlasState.hpp
+++ b/core/AtlasState.hpp
@@ -44,6 +44,7 @@ namespace atlas
     class Translate;
     class Exception;
     class SimController;
+    class VectorState;
 
     using MavisType =
         Mavis<AtlasInst, AtlasExtractor, AtlasInstAllocatorWrapper<AtlasInstAllocator>,
@@ -73,7 +74,7 @@ namespace atlas
         AtlasState(sparta::TreeNode* core_node, const AtlasStateParameters* p);
 
         // Not default -- defined in source file to reduce massive inlining
-        ~AtlasState();
+        virtual ~AtlasState();
 
         HartId getHartId() const { return hart_id_; }
 
@@ -157,6 +158,10 @@ namespace atlas
         const SimState* getSimState() const { return &sim_state_; }
 
         SimState* getSimState() { return &sim_state_; }
+
+        const VectorState* getVectorState() const { return vector_state_ptr_; }
+
+        VectorState* getVectorState() { return vector_state_ptr_; }
 
         const AtlasInstPtr & getCurrentInst() { return sim_state_.current_inst; }
 
@@ -354,6 +359,9 @@ namespace atlas
         //! Simulation state
         SimState sim_state_;
 
+        //! Vector state
+        VectorState* vector_state_ptr_ = nullptr;
+
         // Increment PC Action
         ActionGroup* incrementPc_(AtlasState* state);
         atlas::Action increment_pc_action_;
@@ -434,17 +442,15 @@ namespace atlas
         state->getFpRegister(reg_ident)->dmiWrite<XLEN>(reg_value);
     }
 
-    template <typename XLEN> static inline XLEN READ_VEC_REG(AtlasState* state, uint32_t reg_ident)
+    template <typename VLEN> static inline VLEN READ_VEC_REG(AtlasState* state, uint32_t reg_ident)
     {
-        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
-        return state->getVecRegister(reg_ident)->dmiRead<XLEN>();
+        return state->getVecRegister(reg_ident)->dmiRead<VLEN>();
     }
 
-    template <typename XLEN>
-    static inline void WRITE_VEC_REG(AtlasState* state, uint32_t reg_ident, uint64_t reg_value)
+    template <typename VLEN>
+    static inline void WRITE_VEC_REG(AtlasState* state, uint32_t reg_ident, VLEN reg_value)
     {
-        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
-        state->getVecRegister(reg_ident)->dmiWrite<XLEN>(reg_value);
+        state->getVecRegister(reg_ident)->dmiWrite<VLEN>(reg_value);
     }
 
     template <typename XLEN> static inline XLEN READ_CSR_REG(AtlasState* state, uint32_t reg_ident)
--- a/core/VectorState.hpp
+++ b/core/VectorState.hpp
@@ -1,0 +1,120 @@
+#pragma once
+
+#include <stdint.h>
+#include <iostream>
+
+#include "core/AtlasState.hpp"
+
+namespace atlas
+{
+    class VectorState
+    {
+      public:
+        uint8_t getLMUL() const { return lmul_; }
+
+        uint8_t getSEW() const { return sew_; }
+
+        bool getVTA() const { return vta_; }
+
+        bool getVMA() const { return vma_; }
+
+        uint8_t getVL() const { return vl_; }
+
+        uint8_t getVSTART() const { return vstart_; }
+
+        uint8_t getIndex() const { return index_; }
+
+        template <typename VLEN> inline uint8_t getVLMAX()
+        {
+            return sizeof(VLEN) * lmul_ / (sew_ == 0 ? 1 : sew_);
+        }
+
+        template <typename XLEN> void vsetVTYPE(AtlasState* state, XLEN vtype)
+        {
+            WRITE_CSR_REG<XLEN>(state, VTYPE, vtype);
+            sew_ = 8u << READ_CSR_FIELD<XLEN>(state, VTYPE, "vsew");
+            uint8_t vlmul_ = READ_CSR_FIELD<XLEN>(state, VTYPE, "vlmul");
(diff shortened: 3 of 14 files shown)


core/AtlasInst.hpp
```diff
@@ -115,6 +115,19 @@ namespace atlas
...
+        uint64_t getAVL() const
+        {
+            try
+            {
+                return opcode_info_->getSpecialField(mavis::OpcodeInfo::SpecialField::AVL);
+            }
+            catch (...)
+            {
+                sparta_assert(false, "Can't get SpecialField AVL! " << *this);
+            }
```
knute-mips, line 127: The sparta framework will catch the throw for you.  You don't need to add special code to catch it.

core/inst_handlers/v/RvvInsts.cpp
```diff
@@ -0,0 +1,144 @@
...
+                break;
+            case 0b001:
+                vector_state->lmul = 16;
+                break;
+            case 0b010:
+                vector_state->lmul = 32;
+                break;
+            case 0b011:
+                vector_state->lmul = 64;
+                break;
```
knute-mips, line 70: Is this right?  I'm looking at the spec, and I see that 0b1xx represents fractional LMULs...
Maybe I'm missing how you are representing LMUL here.

lzhu-mips: I store lmul in unit of 1/8. So an integer value can be stored.

core/AtlasState.hpp
```diff
@@ -345,6 +353,9 @@ namespace atlas
         //! Simulation state
         SimState sim_state_;
 
+        //! Vecotr state
```
colby-nyce, line 356: Typo

core/AtlasState.hpp
```diff
@@ -152,6 +153,13 @@ namespace atlas
 
         SimState* getSimState() { return &sim_state_; }
 
+        const VectorState* getVectorState() const
+        {
+            return &vector_state_;
+        } // TODO: what's the point making it private?
```
colby-nyce, line 159: Please remove this TODO comment - public getters are the default design for clean code, not direct member var access. Future-proofing, class invariants, decoupling, caching... many reasons to hide member var access behind an API.

core/VectorState.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include <stdint.h>
+#include <iostream>
+
+namespace atlas
+{
+    struct VectorState
+    {
+        // vtype
+        uint8_t lmul = 8; // unit: one 8th
```
colby-nyce, line 11: What is the reason to use a struct with all these public fields, instead of a regular class with getters (and setters if you need them)?

core/VectorState.hpp
```diff
@@ -0,0 +1,28 @@
...
+        bool vta = false;
+        bool vma = false;
+        // vl
+        uint8_t vl = 0;
+        // vstart
+        uint8_t vstart = 0;
+
+        uint8_t index = 0;
+
+        template <typename VLEN> uint8_t getVLMAX() { return sizeof(VLEN) * lmul / sew; }
```
colby-nyce, line 22: It looks like this can easily hit a divide-by-zero since sew defaults to 0.

core/VectorState.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+    }; // struct VectorState
+
+    std::ostream & operator<<(std::ostream & os, const VectorState & vector_state);
```
colby-nyce, line 26: I would remove VectorState.cpp entirely and inline the stream operator in the header (this is most typical for stream operators -- seems unnecessary to have a .cpp just for this).

core/inst_handlers/v/RvvInsts.cpp
```diff
@@ -0,0 +1,144 @@
...
+        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
+        (void)inst_handlers;
+    }
+
+    template <typename XLEN>
+    void RvvInsts::getInstHandlers(std::map<std::string, Action> & inst_handlers)
+    {
+        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
+
+        // TODO: decide VLEN
```
colby-nyce, line 21: Please expand this comment - it's not clear what the TODO is actually about.

core/inst_handlers/v/RvvInsts.hpp
```diff
@@ -0,0 +1,38 @@
...
+        using D = uint64_t;
+
+        template <typename XLEN>
+        static void getInstComputeAddressHandlers(std::map<std::string, Action> & inst_handlers);
+        template <typename XLEN>
+        static void getInstHandlers(std::map<std::string, Action> & inst_handlers);
+
+      private:
+        template <typename XLEN, typename VLEN>
+        ActionGroup* vsetvl_handler(atlas::AtlasState* state);
```
colby-nyce, line 28: All private functions should append "_" just like all private class member vars

core/VectorState.hpp
lzhu-mips: this is to avoid linking multiple definition error. VectorState.hpp is referenced by AtlasState.hpp, which is used in multiple trans units.

core/inst_handlers/v/RvvInsts.hpp
lzhu-mips: agree, but given current handlers from other extension, I'd keep this in the same form as others.

core/VectorState.hpp
lzhu-mips: this is a very simple aggregate type.  In term of any `state`, it should be easy to initialize.

core/VectorState.cpp
```diff
@@ -0,0 +1,26 @@
...
+            os << "1/" << 8 / vector_state.lmul << " ";
+        }
+        else
+        {
+            os << vector_state.lmul / 8 << " ";
+        }
+        os << "SEW: " << vector_state.sew << " ";
+        os << "VTA: " << vector_state.vta << " "
+           << "VMA: " << vector_state.vma << " ";
+        os << "VL: " << vector_state.vl << " "
```
kathlenemagnus-mips, line 19: Would be good to print VLMAX as well.

core/VectorState.hpp
kathlenemagnus-mips: What about using a float type for LMUL to support fractional LMULs?

kathlenemagnus-mips: I would also recommend a class instead of a struct so we can test whether the values are legal.

core/inst_handlers/v/RvvInsts.hpp
```diff
@@ -0,0 +1,38 @@
...
+    class AtlasState;
+    class Action;
+    class ActionGroup;
+
+    class RvvInsts
+    {
+      public:
+        using base_type = RvvInsts;
+
+        using W = uint32_t;
```
kathlenemagnus-mips, line 18: We have these same using statements in other classes. We should just define them in `AtlasTypes.hpp` instead.

core/inst_handlers/v/RvvInsts.hpp
kathlenemagnus-mips: I can take care of this clean up in another PR. The names of our instruction handlers has been bugging me for awhile!

core/inst_handlers/v/RvvInsts.cpp
```diff
@@ -0,0 +1,144 @@
...
+            atlas::Action::createAction<&RvvInsts::vsetivli_handler<XLEN, VLEN>, RvvInsts>(
+                nullptr, "vsetivli", ActionTags::EXECUTE_TAG));
+    }
+
+    template void RvvInsts::getInstComputeAddressHandlers<RV32>(std::map<std::string, Action> &);
+    template void RvvInsts::getInstComputeAddressHandlers<RV64>(std::map<std::string, Action> &);
+    template void RvvInsts::getInstHandlers<RV32>(std::map<std::string, Action> &);
+    template void RvvInsts::getInstHandlers<RV64>(std::map<std::string, Action> &);
+
+    template <typename XLEN> void RvvInsts::vsetVTYPE(atlas::AtlasState* state, XLEN vtype)
```
kathlenemagnus-mips, line 42: This method and the `vsetAVL` method make more sense as functions of the `VectorState` class.

core/VectorState.hpp
lzhu-pub: About store lmul as float, @knute-mips mentioned that in the comment as well. The reason I keep it as fix-point integer value is 1. avoiding float calculations. 2. ease of printing in form of 1/X. Let me know if this is legit to keep it integer.

core/VectorState.hpp
```diff
@@ -0,0 +1,120 @@
+#pragma once
+
+#include <stdint.h>
+#include <iostream>
+
+#include "core/AtlasState.hpp"
```
kathlenemagnus, line 6: You don't need to fix this in this PR, but I would move the implementation of the methods that require the `AtlasState` definition into a source file. Then, you don't need to include `AtlasState.hpp` in your header; you can just forward declare `AtlasState` instead. This will be a little cleaner since `AtlasState` has to included `VectorState`.

test/sim/vset/Vset_test.cpp
```diff
@@ -20,6 +20,32 @@ class VsetInstructionTester : public AtlasInstructionTester
...
+        const atlas::AtlasState::SimState* sim_state = state->getSimState();
+
+        std::cout << sim_state->current_inst << std::endl;
+        EXPECT_EQUAL(sim_state->inst_count, 2);
+    }
+
+    void testVsetivli()
+    {
+        const uint64_t pc = 0x1000;
+        const uint32_t opcode = 0xc0007057;
```
kathlenemagnus, line 39: Are you planning to check `VectorState` after executing the vset instructions? All would be good to add a comment to explain the value of this opcode.

core/VectorState.hpp
lzhu-pub: it took me a while to settle down to current solution.
`VectorState` is templated and used by different translation unit, so header file has definitions.
If header has definition, then `AtlasState` header is needed, not just forward declaration.
Only pointer of `VectorState` is used in `AtlasState.hpp`.

PR #120, opened 2025-04-15
Hook Up Vector Extension

kathlenemagnus:
Connected all the RVV1.0 instructions to the nop handler to help get @lzhu-mips started with implementing the vset instructions.


klingaard (approved):

kathlenemagnus:
Mavis needs a fix to allow vector to be enabled for RV32. I'll wait on merging this until the Mavis fix is in.

scripts/insts/RVZVE64X_INST.py
```diff
@@ -0,0 +1,15 @@
+RVZVE64X_INST = [
+    {'mnemonic': 'vle64.v', 'handler': 'nop', 'cost': 1, 'tags': 'V_EXT_64', 'memory': True, 'cof': False},
```
lzhu-mips, line 2: how are these matching vset{i}vl{i}?

kathlenemagnus: In `RVZVE32X.py` there are entries for `vsetvl`, `vsetvli`, and `vsetivli`. The handler field will hook these instructions up to the nop handler defined in the `RviInsts` class. This is just a placeholder until @lzhu-pub defines the vector instruction handlers.

kathlenemagnus:
> Mavis needs a fix to allow vector to be enabled for RV32. I'll wait on merging this until the Mavis fix is in.

Updated Mavis with the fix.

PR #121, opened 2025-04-15
Fix for RV32UI tests failure using Instruction handler templates

skhanx86:
Fix for the following tests with update Instruction handlers to template

- rv32ui-p-srl
- rv32ui-p-sra 
- rv32ui-p-srai
- rv32ui-p-slti
- rv32ui-p-srli
- rv32ui-p-slt


core/inst_handlers/i/RviInsts.cpp
```diff
@@ -751,12 +750,12 @@ namespace atlas
...
     }
 
-    ActionGroup* RviInsts::srai_handler(atlas::AtlasState* state)
+    template <typename XLEN> ActionGroup* RviInsts::srai_handler(atlas::AtlasState* state)
     {
         const AtlasInstPtr & insn = state->getCurrentInst();
 
         // require(SHAMT < state->getXlen());
-        const int64_t rs1_val = READ_INT_REG<uint64_t>(state, insn->getRs1());
+        const XLEN rs1_val = READ_INT_REG<uint64_t>(state, insn->getRs1());
```
kathlenemagnus, line 758: Replace template parameter of the register macro with `XLEN` to avoid an unnecessary cast from `uint64_t` to `XLEN`.

core/inst_handlers/i/RviInsts.cpp
```diff
@@ -274,12 +274,12 @@ namespace atlas
...
                             &RviInsts::integer_reg_reg_handler<RV64, std::less<RV64>>, RviInsts>(
                             nullptr, "sltu", ActionTags::EXECUTE_TAG));
-            inst_handlers.emplace("sra",
-                                  atlas::Action::createAction<&RviInsts::sra_handler, RviInsts>(
-                                      nullptr, "sra", ActionTags::EXECUTE_TAG));
-            inst_handlers.emplace("srai",
-                                  atlas::Action::createAction<&RviInsts::srai_handler, RviInsts>(
-                                      nullptr, "srai", ActionTags::EXECUTE_TAG));
+            inst_handlers.emplace(
+                "sra", atlas::Action::createAction<&RviInsts::sra_handler<int64_t>, RviInsts>(
```
lzhu-pub, line 278: use RV64/RV32 instead.

kathlenemagnus: In this case, he wants to use the signed 64-bit type so it's intentional.

core/inst_handlers/i/RviInsts.cpp
skhanx86: When we replace the READ_INT_REG<uint64_t> to READ_INT_REG<XLEN> where XLEN=int32_t or int64_t, it fails due to the assertion in READ_INT_REG.
`    
template <typename XLEN> static inline XLEN READ_INT_REG(AtlasState* state, uint32_t reg_ident)
    {
        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
        return (reg_ident == 0) ? 0 : state->getIntRegister(reg_ident)->dmiRead<XLEN>();
    }
`

core/inst_handlers/i/RviInsts.cpp
skhanx86: Correct, It's intentionally set to signed int types, hence did not use RV64/32, let me know if we should create macros for signed types.

core/inst_handlers/i/RviInsts.cpp
```diff
@@ -751,12 +750,12 @@ namespace atlas
...
-    ActionGroup* RviInsts::srai_handler(atlas::AtlasState* state)
+    template <typename XLEN> ActionGroup* RviInsts::srai_handler(atlas::AtlasState* state)
     {
         const AtlasInstPtr & insn = state->getCurrentInst();
 
         // require(SHAMT < state->getXlen());
-        const int64_t rs1_val = READ_INT_REG<uint64_t>(state, insn->getRs1());
+        const XLEN rs1_val = READ_INT_REG<uint64_t>(state, insn->getRs1());
         const uint64_t shift_amount = insn->getImmediate() & (state->getXlen() - 1);
         const int64_t rd_val = (int64_t)(rs1_val >> shift_amount);
```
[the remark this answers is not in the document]
skhanx86: Done!

core/inst_handlers/i/RviInsts.cpp
kathlenemagnus: In that case, I would recommend setting XLEN back to the unsigned type (`uint32_t1 or `uint64_t`) and using `std::make_signed_t<XLEN>` where you need to use the signed type.

skhanx86: Done!

kathlenemagnus (approved):

PR #123, opened 2025-04-15
Use Instruction Handler Name Instead of Mnemonic

kathlenemagnus:
In `AtlasExtractor`, an instruction's Action Group is initialized with its instruction handler Action(s). The instruction's instruction handler name, defined in the Atlas uarch JSON files, should be used to find the instruction handler Action - NOT the instruction's mnemonic. Fixed this bug and then cleaned up the instruction handler names in the JSONs.


kathlenemagnus:
Thank you @lzhu-pub for pointing out this issue!!!

klingaard (approved):

PR #127, opened 2025-04-24
v ext: add vadd instructions

lzhu-pub:
Work in Progress.
Feedback is welcome.


core/inst_handlers/v/RvviaInsts.cpp
```diff
@@ -0,0 +1,110 @@
...
+                    }
+                    uint16_t result = OP<uint16_t>()(
+                        *(reinterpret_cast<uint16_t*>(rs1_val.data() + (index % (vlenb / sewb)) * sewb)),
+                        *(reinterpret_cast<uint16_t*>(rs2_val.data() + (index % (vlenb / sewb)) * sewb))
+                    );
+                    memcpy(rd_val.data() + (index % (vlenb / sewb)) * sewb, &result, sewb);
+                    if (index == vector_state->getVL() - 1 || index % (vlenb / sewb) == (vlenb / sewb) - 1)
+                    {
+                        WRITE_VEC_REG<VLEN>(state, inst->getRd() + index / (vlenb / sewb), rd_val);
+                    }
```
knute-mips, line 82: Wonder if a `viavv_helper<typename SEW_T>(....args)` is in order

lzhu-pub: Yes!

core/VectorState.hpp
```diff
@@ -10,19 +10,39 @@ namespace atlas
...
+        inline bool getVTA() const { return vta_; }
 
-        bool getVMA() const { return vma_; }
+        inline bool getVMA() const { return vma_; }
 
-        uint8_t getVL() const { return vl_; }
+        inline uint8_t getVL() const { return vl_; }
 
-        uint8_t getVSTART() const { return vstart_; }
+        inline uint8_t getVSTART() const { return vstart_; }
```
klingaard, line 23: There are no guarantees that adding `inline` will make a difference.  Most of the time the compiler will do it for you, so adding it to the header is superfluous.  No harm in adding it, but no advantage either.

core/VectorState.hpp
```diff
@@ -10,19 +10,39 @@ namespace atlas
...
 
-        uint8_t getIndex() const { return index_; }
+        template <uint8_t SEW> struct GetSewType;
+
+        template <> struct GetSewType<8>
```
colby-nyce, line 27: Consider creating a utility function in AtlasUtils.hpp, or even AtlasTypes.hpp:

```
#include <type_traits>

template <std::size_t N>
inline constexpr auto get_uint_type() {
    if constexpr (N == 1) return std::type_identity<uint8_t>{};
    else if constexpr (N == 2) return std::type_identity<uint16_t>{};
    else if constexpr (N == 4) return std::type_identity<uint32_t>{};
    else if constexpr (N == 8) return std::type_identity<uint64_t>{};
    else static_assert(N == 1 || N == 2 || N == 4 || N == 8, "Unsupported size");
}

template <std::size_t N>
using GetUIntType = typename decltype(get_uint_type<N>())::type;
```

You could expand the templates to include "bool Unsigned" if you wanted to make it generic.

core/inst_handlers/v/RvviaInsts.cpp
```diff
@@ -0,0 +1,101 @@
...
+                viavv_helper<VLEN, uint16_t, OP<uint16_t>>(state);
+                break;
+            case 32:
+                viavv_helper<VLEN, uint32_t, OP<uint32_t>>(state);
+                break;
+            case 64:
+                viavv_helper<VLEN, uint64_t, OP<uint64_t>>(state);
+                break;
+            default:
+                assert(false);
```
colby-nyce, line 81: Should this be a sparta_assert? Or an exception? This will compile away in release mode.

core/VectorState.hpp
lzhu-pub: good point.

core/inst_handlers/v/RvviaInsts.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    template void RvviaInsts::getInstHandlers<RV32>(std::map<std::string, Action> &);
+    template void RvviaInsts::getInstHandlers<RV64>(std::map<std::string, Action> &);
+
+    template <typename VLEN, typename SEW, typename OP> void viavv_helper(AtlasState* state)
+    {
+        const AtlasInstPtr & inst = state->getCurrentInst();
+        VectorState* vector_state = state->getVectorState();
+        uint8_t vlenb = sizeof(VLEN);
+        uint8_t sewb = sizeof(SEW);
```
knute-mips, line 40: Can be `const`

core/inst_handlers/v/RvviaInsts.cpp
```diff
@@ -0,0 +1,100 @@
...
+                rs2_val = READ_VEC_REG<VLEN>(state, inst->getRs2() + index / (vlenb / sewb));
+                rd_val = READ_VEC_REG<VLEN>(state, inst->getRd() + index / (vlenb / sewb));
+            }
+            uint8_t result = OP()(
+                *(reinterpret_cast<uint8_t*>(rs1_val.data() + (index % (vlenb / sewb)) * sewb)),
+                *(reinterpret_cast<uint8_t*>(rs2_val.data() + (index % (vlenb / sewb)) * sewb)));
+            memcpy(rd_val.data() + (index % (vlenb / sewb)) * sewb, &result, sewb);
+            if (index == vector_state->getVL() - 1 || index % (vlenb / sewb) == (vlenb / sewb) - 1)
+            {
+                WRITE_VEC_REG<VLEN>(state, inst->getRd() + index / (vlenb / sewb), rd_val);
```
knute-mips, line 56: For performance consider caching `inst->getRsX()`

lzhu-pub:
Ready to merge.

lzhu-pub:
I am curious why my clang-format picked up a bunch of changes. Is it version dependent?

kathlenemagnus (approved):
LGTM. Great testing for the vadd!

kathlenemagnus:
> I am curious why my clang-format picked up a bunch of changes. Is it version dependent?

Yes, there are some differences between different clangformat versions. I wanted to require users to use a specific version, but there isn't 1 version that is supported across all of the platforms we're using for development.

PR #128, opened 2025-04-25
Immediate value calculation fix

klingaard:
Mavis already calculates the immediate value for you, so just use that.

PASS RATE: 395/408
EXPECTED RATE: 395/408

This bug fix also enables other testing.


lzhu-mips (approved):

core/AtlasInst.cpp
```diff
@@ -46,12 +46,18 @@ namespace atlas
         return nullptr;
     }
 
+    uint64_t getImmediateValue(const mavis::OpcodeInfo::PtrType & opcode_info) {
```
colby-nyce, line 49: I would make this private and const, or public and static. And I am not a fan of APIs taking smart pointers like this when it's not needed (i.e. it is not an ownership concern). I would take a const ref.

colby-nyce (approved):

core/AtlasInst.cpp
klingaard: Eh?  This is technically a helper method outside of any class namespace.  It's only visible to AtlasInst.cpp and used by the constructor -- and only once.  I don't see a value in making part of the class...

core/AtlasInst.cpp
```diff
@@ -46,12 +46,18 @@ namespace atlas
...
+        return opcode_info->getImmediateType() == mavis::ImmediateType::SIGNED ?
```
kathlenemagnus-mips, line 50: I'm confused about how this works. Most instructions don't have an immediate value. I thought Mavis asserted if you tried to access an immediate value that doesn't exist?

kathlenemagnus-mips:
Nice clean up! Thanks!

core/AtlasInst.cpp
knute-mips: It actually doesn't, but we could make this an `std::optional` or something...

klingaard: Chatted on the side with Kathlene -- we will always provide an immediate value via the Instruction and trust that the handlers will know when to get that value based on the instruction need.

PR #130, opened 2025-04-29
Support Misaligned Instruction and Data Addresses

### LINKED ISSUE #96: Add Support for Misaligned Memory Accesses
Sparta requires that memory accesses do not cross blocks. Atlas defines the block size as 4K so accesses that cross a 4K boundary need to be broken up into 2 accesses. This applies to both fetch and instruction accesses.

kathlenemagnus-mips:
Added proper misalignment support in Fetch and Translate. For instruction translation, translate will only translate the first part of the misaligned address and then return to Fetch so that Fetch can determine whether the second part of the misaligned address needs to be translated (see detailed comment added to `Fetch::decode_`). For data translation, translate will execute the Translation ActionGroup twice to resolve both translation requests (2B each). 

Also included in this PR:
- Reworked how translation requests and results are handled to support misalignment and instructions with multiple translation requests (groundwork for planned future extension support).
- When an mret or sret instruction sets the next PC, require that the PC to be either 2B (if compression is enabled) or 4B aligned.
- When updating the MISA CSR, disabling the compression extension will fail if doing so would cause a PC misalignment exception.
- When translating misaligned addresses (2x2B requests), if the PTE found during the page table walk is a superpage (larger than 4K), then both translation requests can be resolved at the same time.
- Renamed `insn` variables to `inst` for consistency.

Closes #96


core/translate/Translate.cpp
```diff
@@ -280,15 +280,50 @@ namespace atlas
...
                 const auto index_bits = (vpn_field.msb - vpn_field.lsb + 1) * indexed_level;
                 const auto virt_base = vaddr >> PAGESHIFT;
                 Addr paddr = (Addr(pte.getPpn()) | (virt_base & ((0b1 << index_bits) - 1)))
                              << PAGESHIFT;
                 paddr |= extractPageOffset_(vaddr); // Add the page offset
 
-                translation_state->setResult(paddr, request.getSize());
+                // If there are multiple requests and the page size is larger than 4K, then
+                // multiple requests may be able to be resolved with the current page.
+                if ((translation_state->getNumRequests() > 0) && (page_size != PageSize::SIZE_4K))
```
lzhu-pub, line 294: what if the second request is also crossing the super page boundary?

core/translate/AtlasTranslationState.hpp
```diff
@@ -11,81 +11,101 @@ namespace atlas
...
         }
 
       private:
         // Translation request
-        TranslationRequest request_;
+        std::queue<TranslationRequest> requests_;
 
         // Translation result
-        TranslationResult result_;
+        std::queue<TranslationResult> results_;
```
klingaard, line 109: A cheaper way of doing this: make this an array of requests (max == 2), with a counter for the number of requests.  `pop` would just be a decrement.

klingaard (approved):
I think another approach to this is to assume the translation request is not misaligned from fetch's/LS's POV.  Only when the page is translated, _then_ you can determine that there are two distinct **PAs** for the access. This allows pages > 4K to not "see" a page cross and we only need to do 1 access.

core/translate/AtlasTranslationState.hpp
kathlenemagnus-mips: If I move the misalignment detection to Translate then I may just need to support a single request anyways.

core/translate/Translate.cpp
kathlenemagnus-mips: I hadn't considered that scenario. I will add a "isContained" method to the PTE so make sure that both requests can be handled by the page.

kathlenemagnus-mips:
I realized that I am handling superpage translation incorrectly. I need to rework how the offset and PPN are determined to support superpages properly. I'm going to close this PR for now and will open a new PR with the fixes.

PR #132, opened 2025-05-05
Support Misaligned Instruction and Data Addresses (Attempt #2)

kathlenemagnus-mips:
Added proper misalignment support in Fetch and Translate. Misalignment detection is performed at the end of translation so that the current page size can be considered. If misalignment is detected, a new translation request is made by Translate. For instruction translation, translate will only translate the first part of the misaligned address and then return to Fetch so that Fetch can determine whether the second part of the misaligned address needs to be translated (see detailed comment added to Fetch::decode_). For data translation, translate will execute the Translation ActionGroup twice to resolve both translation requests (2B each).

Also included in this PR:
- When an mret or sret instruction sets the next PC, require that the PC to be either 2B (if compression is enabled) or 4B aligned.
- When updating the MISA CSR, disabling the compression extension will fail if doing so would cause a PC misalignment exception.
- Renamed insn variables to inst for consistency.


kathlenemagnus-mips:
FYI I did not include support for multiple translation requests for vector in this PR. I'll work on that next.

core/Fetch.cpp
```diff
@@ -60,23 +61,75 @@ namespace atlas
     ActionGroup* Fetch::decode_(AtlasState* state)
     {
         // Get translation result
-        const AtlasTranslationState::TranslationResult & result =
+        const AtlasTranslationState::TranslationResult result =
```
lzhu-pub, line 64: what is the reason to remove reference?

lzhu-pub (approved):

PR #137, opened 2025-05-08
v ext: refactoring vcs and via tests

lzhu-pub:
cleared the boundary of vcs anv via tests.
used mavis to construct AtlasInst for tests.


klingaard (approved):
Awesome tests!

PR #138, opened 2025-05-09
Added Command Line Option for Overriding a Register's Initial Value

kathlenemagnus-mips:
Also added support for RD2 in AtlasInst.

### DIFF
--- a/sim/atlas.cpp
+++ b/sim/atlas.cpp
@@ -2,7 +2,7 @@
 #include "sparta/app/CommandLineSimulator.hpp"
 
 const char USAGE[] = "Usage:\n"
-                     "./atlas [-i inst limit] <workload>"
+                     "./atlas [-i inst limit] [--reg \"name value\"] [--interactive] [--spike-formatting] <workload>"
                      "\n";
 
 struct regOverride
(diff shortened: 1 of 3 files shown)


sim/atlas.cpp
```diff
@@ -79,6 +96,31 @@ int main(int argc, char** argv)
...
+        const atlas::HartId hart_id = 0;
+        atlas::AtlasState* state = sim.getAtlasState(hart_id);
+        if (vm.count("reg"))
+        {
+            auto & reg_overrides = vm["reg"].as<std::vector<regOverride>>();
+            for (auto reg_override : reg_overrides)
+            {
+                bool hex_value = reg_override.value.find("0x", 0) != std::string::npos;
+                const uint64_t reg_value =
+                    std::strtoull(reg_override.value.c_str(), nullptr, (hex_value ? 16 : 10));
```
klingaard, line 108: Consider using `std::stoll` with a base of `0` which means to interpret the string as hex, octal, etc:
https://en.cppreference.com/w/cpp/string/basic_string/stol

klingaard (approved):

PR #139, opened 2025-05-13
Redesign the Action and ActionGroup Classes

kathlenemagnus-mips:
To support multiple translation requests, the Translate Actions in an instruction's ActionGroup need to be executed more than once. The current ActionGroup design does not support this, because an executing Action does not have a way to communicate that it needs to be executed again. This PR reworks the ActionGroup "execute" method.


kathlenemagnus-mips:
This PR is still a WIP. Open a draft PR to get input from the rest of the team.

kathlenemagnus:
> This PR is still a WIP. Open a draft PR to get input from the rest of the team.

This PR is ready for review!

core/ActionGroup.hpp
```diff
@@ -56,15 +69,19 @@ namespace atlas
 
         const std::vector<Action> & getActions() const { return action_group_; };
 
-        ActionGroup* execute(AtlasState* state) const
+        ActionGroup* execute(AtlasState* state)
```
kathlenemagnus, line 72: This method is the most important part of this PR.

core/Action.hpp
```diff
@@ -42,6 +42,11 @@ namespace atlas
      */
     class Action
     {
+      public:
+        //! \brief Iterator type
+        typedef std::vector<Action>::iterator ItrType;
```
knute-mips, line 47: Nit.
```suggestion
        using ItrType = std::vector<Action>::iterator;
```

core/ActionGroup.hpp
```diff
@@ -56,15 +69,21 @@ namespace atlas
...
         {
-            ActionGroup* fail_action_group = nullptr;
-            for (const Action & action : action_group_)
+            Action::ItrType action_it = action_group_.begin();
+            while (action_it != action_group_.end())
```
knute-mips, line 75: If the action group will not be changed during the execution of the loop, cache the end iterator so the compiler doesn't have to reload it for every comparison.

core/ActionGroup.cpp
```diff
@@ -3,6 +3,8 @@
 
 namespace atlas
 {
+    const char* ActionException::what() const throw() { return action_group_->getName().c_str(); }
```
lzhu-pub, line 6: use `noexcept` instead

core/ActionGroup.hpp
```diff
@@ -8,6 +8,19 @@
...
 namespace atlas
 {
+    class ActionException : public std::exception
+    {
+      public:
+        ActionException(ActionGroup* action_group) : action_group_(action_group) {}
+
+        ActionGroup* getActionGroup() { return action_group_; }
+
+        const char* what() const throw();
```
lzhu-pub, line 18: `noexcept`

core/ActionGroup.hpp
```diff
@@ -56,15 +69,19 @@ namespace atlas
...
+            fail_action_group_ = nullptr;
+            Action::ItrType action_it = action_group_.begin();
+            while (action_it != action_group_.end())
             {
-                fail_action_group = action.execute(state);
-                if (SPARTA_EXPECT_FALSE(fail_action_group))
+                try
                 {
-                    return fail_action_group;
+                    action_it = action_it->execute(state, action_it);
```
lzhu-pub, line 81: I can slightly feel that passing current action_it to Action Method is better than the next action_it. It might come in handy at some point when Action Method needs its own action_it.

klingaard (approved):

core/ActionGroup.hpp
kathlenemagnus: > I can slightly feel that passing current action_it to Action Method is better than the next action_it. It might come in handy at some point when Action Method needs its own action_it.

@lzhu-pub the current Action is what is passed into the Action method. The next Action is what is returned.

PR #147, opened 2025-05-21
Refactor common code into Observer base class

colby-nyce:


core/observers/Observer.hpp
```diff
@@ -15,28 +14,72 @@ namespace atlas
...
-        Observer() { reset_(); }
+        // The base class needs to know if we are rv32 or rv64 since it is responsible for
+        // reading register values (XLEN).
+        //
+        // We do not use templates here (template <typename XLEN>) because then AtlasState
+        // cannot hold onto arch-agnostic observers (std::vector<std::unique_ptr<Observer>>).
+        //
+        // Note that if your subclass tells the Observer to use Arch::UNUSED, then the
+        // before/after register values will NOT be tracked.
+        enum class Arch
```
kathlenemagnus, line 25: This is a little nit-picky, but can we rename this enum to something like"Mode" or "ObserverMode"? Since this enum is used to determine what is being observed (RV32 state or RV64 state), then the "UNUSED" value makes more sense to me.

kathlenemagnus (approved):
This is GREAT. Love to see so much duplicate code being removed!

core/observers/Observer.hpp
colby-nyce: Done

core/observers/InstructionLogger.cpp
```diff
@@ -117,64 +117,75 @@ namespace atlas
...
+                using SXLEN = int32_t;
+                const SXLEN imm_val = imm;
+                inst_oss_ << "   imm: " << HEX(imm_val, getRegWidth());
+            }
+            else
+            {
+                using SXLEN = int64_t;
+                const SXLEN imm_val = imm;
+                inst_oss_ << "   imm: " << HEX(imm_val, getRegWidth());
+            }
```
knute-mips, line 132: Haven't had coffee yet, but is this swapped?

core/observers/InstructionLogger.cpp
```diff
@@ -117,64 +117,75 @@ namespace atlas
         void writeImmediate(const uint64_t imm) override
         {
             reset_();
-            using SXLEN = typename std::make_signed<XLEN>::type;
-            const SXLEN imm_val = imm;
-            inst_oss_ << "   imm: " << HEX(imm_val, WIDTH);
+
+            if (getRegWidth() == 8)
```
knute-mips, line 121: Might want to do:
```
if (getRegWidth() == sizeof(int64_t))
```

core/observers/Observer.cpp
```diff
@@ -1,8 +1,130 @@
...
+        interrupt_cause_ = state->getExceptionUnit()->getUnhandledInterrupt();
+
+        // Subclass impl
+        preException_(state);
+    }
+
+    void Observer::postExecute(AtlasState* state)
+    {
+        // Get final value of destination registers
+        AtlasInstPtr inst = state->getCurrentInst();
```
knute-mips, line 39: Since `inst` is only used to check for `nullptr`, instead of copying the `inst`, just create a local `const bool inst_valid = state->getCurrentInst();`

core/observers/Observer.cpp
```diff
@@ -1,8 +1,130 @@
...
+        }
+
+        // Subclass impl
+        postExecute_(state);
+    }
+
+    void Observer::inspectInitialState_(AtlasState* state)
+    {
+        pc_ = state->getPc();
+        AtlasInstPtr inst = state->getCurrentInst();
```
knute-mips, line 81: One performance suggestion, deference the smart pointer and use the local pointer:
```
AtlasInst * inst = state->getCurrentInst()->get();

PR #148, opened 2025-05-21
v ext: add vlse instruction handlers

lzhu-pub:
Adding vector load store instruction handlers and tests.
1. unit stride load/store
2. strided load/store
3. ordered-indexed load/store


test/sim/vls/Vls_test.cpp
```diff
@@ -0,0 +1,196 @@
...
+        offset += 5;
+        opcode |= 0 << offset; // rs2
+        offset += 5;
+        opcode |= vm << offset; // vm
+        offset += 1;
+        opcode |= 0 << offset; //newop
+        offset += 3;
+        opcode |= 0 << offset; // nf
+        offset += 3;
+        assert(offset == 32);
```
knute-mips, line 130: Probably should use `EXPECT_EQUAL(offset, 32)` instead of an assert.  These asserts won't be caught unless you compile in debug mode.

lzhu-pub: this check is not part of the test. it's a soft check that opcode build function is written correctly.

core/inst_handlers/v/RvvlsInsts.cpp
```diff
@@ -0,0 +1,368 @@
...
+    template <typename XLEN, typename VLEN, typename EEW, RvvlsInsts::Mop mop>
+    Action::ItrType RvvlsInsts::vlseComputeAddressHandler_(atlas::AtlasState* state,
+                                                           Action::ItrType action_it)
+    {
+        static_assert(std::is_same<XLEN, RV64>::value || std::is_same<XLEN, RV32>::value);
+
+        const AtlasInstPtr inst = state->getCurrentInst();
+        VectorState* vector_state = state->getVectorState();
+        const uint8_t vlenb = sizeof(VLEN);
+        const uint8_t eewb = sizeof(EEW);
```
kathlenemagnus, line 241: Instead of using another template parameter for the EEW, can we get this value from the instruction instead? If it's not available from Mavis, we can add it to our Atlas uarch JSONs and get it from the `AtlasExtractor`.

core/inst_handlers/v/RvvlsInsts.cpp
```diff
@@ -0,0 +1,368 @@
...
+            for (; index < vl; ++index)
+            {
+                if (inst->getVM()
+                    || (reinterpret_cast<const uint8_t*>(&mask)[index / 8] >> (index % 8)) & 1)
+                {
+                    inst->getTranslationState()->makeRequest(rs1_val + index * stride, eewb);
+                }
+            }
+        }
+        else // Indexed
```
kathlenemagnus, line 270: I think it would be cleaner to just create a separate compute address handler for the indexed instructions.

core/inst_handlers/v/RvvlsInsts.cpp
```diff
@@ -0,0 +1,368 @@
...
+                                           result);
+                    inst->getTranslationState()->popResult();
+                }
+            }
+        }
+        if constexpr (load)
+        {
+            WRITE_VEC_REG<VLEN>(state, rd + vstart / (vlenb / ewb), rd_val);
+        }
+        vector_state->setVSTART(index);
```
kathlenemagnus, line 335: Vstart should only be set if an exception occurs, which we don't support yet for vector loads and stores. In VectorState, I would add some variables to track the execution of the current vector instruction.

core/inst_handlers/v/RvvlsInsts.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+    template <typename VLEN, bool load>
+    Action::ItrType RvvlsInsts::vlseIdxHandler_(atlas::AtlasState* state, Action::ItrType action_it)
+    {
+        VectorState* vector_state = state->getVectorState();
+        switch (vector_state->getSEW())
+        {
+            case 8:
+                return vlseHandler_<VLEN, uint8_t, load>(state, action_it);
+                break;
```
kathlenemagnus, line 351: I think you can remove the break statements. I would also try to remove the return on line 365 since it should be unreachable.

core/inst_handlers/v/RvvlsInsts.cpp
```diff
@@ -0,0 +1,368 @@
...
+            return action_it;
+        }
+        return ++action_it;
+    }
+
+    template <typename VLEN, bool load>
+    Action::ItrType RvvlsInsts::vlseIdxHandler_(atlas::AtlasState* state, Action::ItrType action_it)
+    {
+        VectorState* vector_state = state->getVectorState();
+        switch (vector_state->getSEW())
```
kathlenemagnus, line 347: We should consider defining enums for these values that have a very limited range of valid values. For LMUL, SEW, EEW, etc.

kathlenemagnus:
I really like where we're going with this design. There's still a lot to do, but this is a great start!

core/inst_handlers/v/RvvlsInsts.cpp
lzhu-pub: Good point. Here is my thought about `AtlasInst`. Ideally it should only contain member variables that is mutable for decoded instruction. Whereas EEW is fixed for any decoded vleX. If we followed the logic to put everything in `AtlasInst` for any used Field, the size of `AtlasInst` would explode.

core/inst_handlers/v/RvvlsInsts.cpp
lzhu-pub: I can feel your concern. Currently `vstart` is just a variable. Since it's not updating `vstart csr`, it has nothing to do with exception yet. In case we add exception support and hook up csr in the future, the value of `vstart` in current implementation would still be a legit value, as the execution has reached the code there and an Exception could occur. It's perfectly fine to not use  `vstart` value to update csr should no exception happens.

core/inst_handlers/v/RvvlsInsts.cpp
lzhu-pub: If we want to safeguard software invalid input, then `vsetVTYPE()` has already done that.
If we want to safeguard coding bugs, then we could add enums and limitations.
Which one is the target here?

core/AtlasState.cpp
```diff
@@ -100,9 +100,9 @@ namespace atlas
     // Not default -- defined in source file to reduce massive inlining
     AtlasState::~AtlasState()
     {
-        if (vector_state_ptr_)
+        if (vector_config_ptr_)
         {
-            delete vector_state_ptr_;
+            delete vector_config_ptr_;
```
kathlenemagnus, line 105: I don't think this is necessary to add to the destructor.

core/AtlasState.hpp
```diff
@@ -431,6 +431,19 @@ namespace atlas
         state->getVecRegister(reg_ident)->dmiWrite<VLEN>(reg_value);
     }
 
+    template <typename Elem>
```
kathlenemagnus, line 434: Instead of `Elem` let's use `EEW` ("effective element width") from the vector spec.

core/VectorState.hpp
```diff
@@ -2,21 +2,49 @@
...
 #include "core/AtlasState.hpp"
+#include "core/VecNums.hpp"
+#include "include/AtlasUtils.hpp"
 
 namespace atlas
 {
-    class VectorState
+    constexpr uint8_t VLEN_MIN = 32;
+
+    class VectorConfig
```
kathlenemagnus, line 15: Rename file to `VectorConfig.hpp` to match the class name.

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
-            return vl_;
+            setVSTART(vstart);
+            WRITE_CSR_REG<XLEN>(state_ptr, VSTART, vstart);
         }
 
       private:
-        // vtype csr
+        // member variables
+
+        uint8_t vlen_ = 64;
```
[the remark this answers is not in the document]
kathlenemagnus: Actually scratch that idea - we should add a constructor to this class to provide the initial values. Then, add parameters to `AtlasState` to set the initial values.

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
-        inline uint8_t getVSTART() const { return vstart_; }
+        inline uint8_t getVSTART() const { return vstart; }
 
-        inline void setVSTART(uint8_t value) { vstart_ = value; }
+        inline void setVSTART(uint8_t value) { vstart = value; }
 
-        template <typename VLEN> inline uint8_t getVLMAX()
+        // helper functions
+
+        inline uint8_t getVLMAX() { return getVLEN() / 8 * getLMUL() / getSEW(); }
```
kathlenemagnus, line 67: Shouldn't this just be:
```
(getVLEN() / getSEW()) * getLMUL()
```
Also the return type is not large enough. I would use `uint64_t`.

core/inst_handlers/v/RvviaInsts.cpp
```diff
@@ -13,81 +11,58 @@ namespace atlas
...
-            memcpy(rd_val.data() + (index % (vlenb / sewb)) * sewb, &result, sewb);
-        }
-        WRITE_VEC_REG<VLEN>(state, rd + vstart / (vlenb / sewb), rd_val);
-        vector_state->setVSTART(index);
-        if (index != vl)
+        VectorElements<ElemWidth> elems_vs1{state, inst->getRs1(), false};
+        VectorElements<ElemWidth> elems_vs2{state, inst->getRs2(), false};
+        VectorElements<ElemWidth> elems_vd{state, inst->getRd(), false};
+
+        for (auto iter_vs1 = elems_vs1.begin(), iter_vs2 = elems_vs2.begin(),
```
kathlenemagnus, line 39: Wow! This code is so much cleaner that the original design! I like it.

core/inst_handlers/v/RvvlsInsts.cpp
```diff
@@ -0,0 +1,287 @@
...
+        VectorElements<ElemWidth> elems_vs2{state, inst->getRs2(), !inst->getVM()};
+        for (auto stride : elems_vs2)
+        {
+            inst->getTranslationState()->makeRequest(rs1_val + stride.getVal(), sewb);
+        }
+
+        return ++action_it;
+    }
+
+    template <uint8_t ElemWidth, bool load>
```
kathlenemagnus, line 239: Change `load` to `LOAD`.

core/inst_handlers/v/RvviaInsts.cpp
```diff
@@ -13,81 +11,58 @@ namespace atlas
...
+    Action::ItrType RvviaInsts::viavvHandler_(AtlasState* state, Action::ItrType action_it)
     {
-        VectorState* vector_state = state->getVectorState();
-        switch (vector_state->getSEW())
+        VectorConfig* vector_config_ptr = state->getVectorConfig();
+        switch (vector_config_ptr->getSEW())
         {
             case 8:
-                return viavv_helper<VLEN, uint8_t, OP<uint8_t>>(state, action_it);
+                return viavvHelper<8, FunctorTemp<uint8_t>>(state, action_it);
```
kathlenemagnus, line 56: If you change the template parameter to a type instead of an integer, you can pass in `uint8_t` instead. That way you don't have to call `GetUintType` to convert from the size to a type.

core/inst_handlers/v/RvviaInsts.cpp
```diff
@@ -13,81 +11,58 @@ namespace atlas
...
+            atlas::Action::createAction<&RvviaInsts::viaviHandler_<std::plus>, RvviaInsts>(
+                nullptr, "vadd.vi", ActionTags::EXECUTE_TAG));
     }
 
     template void RvviaInsts::getInstHandlers<RV32>(std::map<std::string, Action> &);
     template void RvviaInsts::getInstHandlers<RV64>(std::map<std::string, Action> &);
 
-    template <typename VLEN, typename SEW, typename OP>
-    Action::ItrType viavv_helper(AtlasState* state, Action::ItrType action_it)
+    template <uint8_t ElemWidth, typename Functor>
```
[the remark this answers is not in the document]
kathlenemagnus: Just to explain my other comment more... this can be:
```
    template <typename SEW, typename FUNCTOR>
```

core/AtlasState.hpp
lzhu-pub: This has been bugging me for a while. Theoretically `VLEN`, `XLEN`, `SEW`, `EEW` are values, not types. When we `getSEW()`, we get a value. If we really want to be consistent with `XLEN`, then we can just follow it on that path...

core/VectorState.hpp
lzhu-pub: The file also defines other classes. Is it ok to use the first defined class as file name?

lzhu-pub:
> Where is `VectorConfig::setVLEN` called today? Since the vector register size comes from the register definition JSONs in `arch` we need to figure out how to have the user pass in a parameter to specify the size.
> 
> My suggestion would be to generate vector register definition JSONs for ALL supported sizes and add a parameter to `AtlasState` to specify the VLEN. Then, `AtlasState` can select the correct vector register definition JSON. We can then also make VLEN a const value of `VectorConfig`.

it's not called yet. `vlen_` relies on it's initial value and never changes so far. We can add additional support in a separate PR.

core/VectorState.hpp
lzhu-pub: What is the max `VL` we'd like to support? So far I am using the minimum type for all `sew_`, `vl_`  etc..
 I could change then all to uint32_t, or uint64_t.

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
-        uint8_t vstart_ = 0;
-    }; // struct VectorState
 
-    template <typename VLEN>
-    inline std::ostream & operator<<(std::ostream & os, const VectorState & vector_state)
+        // VSTART CSR
+        uint8_t vstart = 0;
+    }; // class VectorConfig
+
+    template <uint8_t ElemWidth> struct VectorElement
```
kathlenemagnus, line 137: What about if this struct just contained a dmi pointer to the register? You could take the dmi pointer to the vector register and then increment it by the "element size * idx" to get a pointer to the specific element in the vector register. We would need to confirm that the vector registers are stored in memory contiguously so that if you increment the dmi pointer by VLEN that you get to the next register.

core/VectorState.hpp
colby-nyce, line 67: Simple getters should be public (getLMUL, getSEW...)

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
+            state_ptr_(state_ptr),
+            config_ptr_(state_ptr_->getVectorConfig()),
+            index_(config_ptr_->getVSTART()),
+            reg_id_(reg_id),
+            masked_(masked),
+            mask_(masked_ ? VectorMask{state_ptr, atlas::V0} : VectorMask{})
+        {
+        }
+
+        inline bool isMasked() { return masked_; }
```
colby-nyce, line 271: Simple getter functions returning member var copies are always const

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
+        class ElementIterator
+        {
+          public:
+            using difference_type = std::ptrdiff_t;
+            using value_type = VectorElement<ElemWidth>;
+            using pointer = VectorElement<ElemWidth>*;
+            using reference = VectorElement<ElemWidth> &;
+            using iterator_category = std::forward_iterator_tag;
+
+            ElementIterator(VectorElements<ElemWidth>* elems_ptr, uint8_t index) :
```
colby-nyce, line 201: Iterator classes typically have private constructors since the "outer" friend class always creates them.

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
+                elems_ptr_(elems_ptr),
+                index_(index)
+            {
+                if (elems_ptr_->masked_)
```
colby-nyce, line 205: Use this logic on the outside (in the call to `begin()` and `end()` and give the correct index directly to the ctor.

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
+
+    template <uint8_t ElemWidth> struct VectorElement
+    {
+        using ElemType = GetUintType<ElemWidth>;
+
+        void setVal(ElemType value) { WRITE_VEC_ELEM<ElemType>(state_ptr, reg_id, value, idx); }
+
+        ElemType getVal() { return READ_VEC_ELEM<ElemType>(state_ptr, reg_id, idx); }
+
+        AtlasState* state_ptr;
```
colby-nyce, line 145: No public member vars. Does this class need a proper constructor? I doubt you can create one without the AtlasState and the reg_id.

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
         }
-        os << "SEW: " << vector_state.getSEW() << " ";
-        os << "VTA: " << vector_state.getVTA() << " "
-           << "VMA: " << vector_state.getVMA() << " ";
-        os << "VL: " << vector_state.getVL() << " "
-           << "VLMAX:" << vector_state.getVLMAX<VLEN>() << " "
-           << "VSTART: " << vector_state.getVSTART() << "; ";
+        os << "SEW: " << config.getSEW() << " ";
+        os << "VTA: " << config.getVTA() << " " << "VMA: " << config.getVMA() << " ";
+        os << "VL: " << config.getVL() << " " << "VSTART: " << config.getVSTART() << "; ";
```
colby-nyce, line 303: Do we want to print out these uint8_t values like this? I know the code was like that before, but still... is it better to cast these to an int? Maybe Kathlene can chime in.

Also, please watch out for std::cout with bools and use std::boolalpha if you want better readability.

core/inst_handlers/v/RvvcsInsts.cpp
```diff
@@ -1,9 +1,8 @@
 #include "core/inst_handlers/v/RvvcsInsts.hpp"
 #include "core/AtlasState.hpp"
-#include "core/VectorState.hpp"
 #include "core/ActionGroup.hpp"
+#include "core/VectorState.hpp"
```
colby-nyce, line 4: Why did you move this line of code? If it won't compile without being included in a certain order, it is likely "code smell" and needs some refactoring/forward declarations/etc.

core/inst_handlers/v/RvviaInsts.cpp
```diff
@@ -1,10 +1,8 @@
 #include "core/inst_handlers/v/RvviaInsts.hpp"
 #include "core/AtlasState.hpp"
-#include "core/VectorState.hpp"
 #include "core/ActionGroup.hpp"
+#include "core/VectorState.hpp"
```
colby-nyce, line 4: Same question about possible code smell

core/inst_handlers/v/RvvlsInsts.cpp
```diff
@@ -0,0 +1,287 @@
...
+    Action::ItrType RvvlsInsts::vlseHandler_(atlas::AtlasState* state, Action::ItrType action_it)
+    {
+        const AtlasInstPtr inst = state->getCurrentInst();
+
+        VectorElements<ElemWidth> elems_vd{state, inst->getRd(), !inst->getVM()};
+        for (auto iter_vd = elems_vd.begin(); iter_vd != elems_vd.end(); ++iter_vd)
```
colby-nyce, line 245: Can you see if your code compiles when you use range-based for loops?

```
for (auto& vd : elems_vd) {
    ...
    vd.setVal(value);
}
```

core/inst_handlers/v/RvvlsInsts.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    class RvvlsInsts
+    {
+      public:
+        using base_type = RvvlsInsts;
+
+        enum struct AddressingMode
+        {
+            Unit,
+            IdxUnordered,
```
colby-nyce, line 23: Enumerations should be all caps

system/AtlasSystem.cpp
```diff
@@ -10,7 +10,8 @@ namespace atlas
         sparta::Unit(sys_node),
         syscall_emulation_enabled_(p->enable_syscall_emulation)
     {
-        if (const std::string workload = p->workload; false == workload.empty()) {
+        if (const std::string workload = p->workload; false == workload.empty())
```
colby-nyce, line 13:
```
if (!p->workload.empty()) {
    loadWorkload_(p->workload);
}
```

system/AtlasSystem.cpp
```diff
@@ -38,8 +39,8 @@ namespace atlas
             if (memory_section.data != nullptr)
             {
                 std::cout << "  -- Loading section " << memory_section.name << " (" << std::dec
-                          << memory_section.file_size << "B) "
-                          << " to 0x" << std::hex << memory_section.start_address << std::endl;
+                          << memory_section.file_size << "B) " << " to 0x" << std::hex
```
colby-nyce, line 42: Be careful to (usually, and probably in this case) go back to `std::dec` if your line of code went to `std::hex` otherwise log files and stdout messages could get messed up and still print hex values when it shouldn't.

test/sim/InstructionTester.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include "sim/AtlasSim.hpp"
 #include "core/AtlasState.hpp"
+#include "core/VectorState.hpp"
```
colby-nyce, line 5: Should not be needed

test/sim/vls/Vls_test.cpp
```diff
@@ -0,0 +1,195 @@
...
+        return opcode;
+    }
+
+  private:
+    atlas::AtlasInst::PtrType instPtr_ = nullptr;
+};
+
+int main()
+{
+    VlsInstructionTester Vls_tester;
```
colby-nyce, line 187: You don't need this class. You could just have free methods `testVle8()` etc. (`instPtr_` is not used)

test/stf/STF_test.cpp
```diff
@@ -12,8 +12,8 @@ int main()
     stf::STFWriter stf_writer;
 
     stf_writer.open("stf_write_test.zstf");
-    stf_writer.addTraceInfo(stf::TraceInfoRecord(stf::STF_GEN::STF_GEN_DROMAJO,
-                                                 1, 2, 0, "Trace from Dromajo"));
+    stf_writer.addTraceInfo(
```
colby-nyce, line 15: Did you run clang-format? This line of code didn't change. There are other places in the PR where it looks like clang-format was skipped.

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
 
-        inline void setVSTART(uint8_t value) { vstart_ = value; }
+        inline void setVSTART(uint8_t value) { vstart = value; }
 
-        template <typename VLEN> inline uint8_t getVLMAX()
+        // helper functions
+
+        inline uint8_t getVLMAX() { return getVLEN() / 8 * getLMUL() / getSEW(); }
+
+        template <typename XLEN> XLEN vsetAVL(AtlasState* state_ptr, bool set_max, XLEN avl = 0)
```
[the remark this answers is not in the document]
lzhu-mips: how about we change all others to `AtlasState* state_ptr`?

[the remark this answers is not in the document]
colby-nyce: Let's keep everything called `state` since it is used everywhere and takes up less space / easier to type

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
+
+            ElementIterator operator++(int)
+            {
+                ++(*this);
+                return *this;
+            }
+
+            bool operator!=(const ElementIterator & other) const
+            {
+                return index_ != other.index_ && elems_ptr_ == other.elems_ptr_;
```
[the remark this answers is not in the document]
lzhu-pub: interestingly after implementing `==` and re-use it in `!=`, I got C++20 error when calling `iter != Elements.end()`:
```
error: ISO C++20 considers use of overloaded operator '!=' (with operand types 'atlas::MaskElements::MaskElementIterator' and 'atlas::MaskElements::MaskElementIterator') to be ambiguous despite there being a unique best viable function [-Werror,-Wambiguous-reversed-operator]
                     mask_elem_iter != MaskElementIterator{elems_ptr_, elems_ptr_->end_pos_};
                     ~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/lzhu/Git/atlas/core/VectorState.hpp:288:18: note: ambiguity is between a regular call to this operator and a call with the argument order reversed
            bool operator==(const IteratorIF & other) const override
```

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
+                    index_ = elems_ptr_->mask_.getNextIndex(++index_);
+                }
+                else
+                {
+                    ++index_;
+                }
+                return *this;
+            }
+
+            ElementIterator operator++(int)
```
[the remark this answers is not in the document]
colby-nyce: You can leave just what you need right now for the iterator class (pre-increment to save on performance).

But I'll give my "PR feedback" for your next PR: using polymorphic iterators would be too bad performance and is almost certainly not needed. The current code will just be tweaked but we still have to use tiny iterator objects and optimized code for the operator++ math. But vtable indirection is many times slower than updating an index. And you would have to "new" an iterator every time the loop hits. I wouldn't use that design IMHO.

core/VectorState.hpp
lzhu-pub: I'd love to have such dmi pointer if sparta could provide it! `raw_data_ptr_ ` in `Register` seems not accessible.

core/inst_handlers/v/RvvcsInsts.cpp
lzhu-pub: not intentionally, maybe I deleted it, tried something else, didn't work, and moved it back, at a different location.

core/AtlasState.cpp
lzhu-pub: Maybe I should change everything to be `shared_ptr`? I was looking at sparta smart pointers but i failed to understand why we have it instead of smart pointers.

core/VectorState.hpp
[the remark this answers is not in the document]
lzhu-pub: I have added constructor with valid list initialization in upcoming PR.

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
+                }
+                ++index;
+            }
+            return index;
+        }
+
+      private:
+        AtlasState* state_ptr_;
+        const VectorConfig* config_ptr_;
+        std::vector<GetUintType<VLEN_MIN>> mask_;
```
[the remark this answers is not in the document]
lzhu-pub: I will rename it to `UintType`. I'd like to keep PascalCase to distinguish it from std typenames.

core/VectorState.hpp
lzhu-pub: It can be done for `ElementIterator`, but for upcoming PR, there is class `MaskElements`, with both `MaskElementIterator` and `MaskBitIterator`. `begin()` cannot differentiate those 2 iterator selections. So public constructor is preserved.
Also, public constructor has `index` parameter. It serves special purpose, which might be of use. We can discuss it in later PR>

core/VectorState.hpp
lzhu-pub: This issue will be harder to solve when polymorphism is involved. Let's discuss this later in next PR.

core/VectorState.hpp
```diff
@@ -30,89 +58,249 @@ namespace atlas
...
+            {
+                if (elems_ptr_->masked_)
+                {
+                    index_ = elems_ptr->mask_.getNextIndex(index_);
+                }
+            }
+
+            reference operator*()
+            {
+                elems_ptr_->current_.state_ptr = elems_ptr_->state_ptr_;
```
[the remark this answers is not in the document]
lzhu-pub: On the contrary, it is to improve performance to execute `Element` update only when we want to get the `Element`. We don't want to update `current_` each time we increment iterator.

system/AtlasSystem.cpp
lzhu-pub: not sure why i have this change. clangformat...
but i will fix this.

test/stf/STF_test.cpp
lzhu-pub: yes i did. it seems that different clang-format gives different result.

test/sim/vls/Vls_test.cpp
lzhu-pub: do you mean to make `testVle8` `static`?

PR #159, opened 2025-06-03
Implemented the ZBA Instructions and most of the available ZBB Instructions

DeclanNKim:
There were some instructions that were either missing or extraneous as a result of the RISC-V Instruction Set Manual updating fairly recently, so those will be implemented in the future. Additionally, I was unable to properly test the RORI and RORIW instructions due to an outside error.


lzhu-pub:
please remove the test log files.

PR #161, opened 2025-06-04
Implemented B-Extension Instructions originally available in Atlas

DeclanNKim:
Since Mavis was using an older version of the B extension, I only implemented the instructions that overlapped between the most up-to-date Instruction Set Manual specs and Mavis.


core/inst_handlers/b/RvzbaInsts.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    template void RvzbaInsts::getInstHandlers<RV32>(std::map<std::string, Action> &);
+    template void RvzbaInsts::getInstHandlers<RV64>(std::map<std::string, Action> &);
+
+    Action::ItrType RvzbaInsts::add_uw_handler(atlas::AtlasState* state, Action::ItrType action_it)
+    {
+        const AtlasInstPtr & inst = state->getCurrentInst();
+
+        // Ensures that rs1_val only receives the least-significant word of rs1
+        const uint64_t rs1_val = static_cast<uint64_t>(static_cast<uint32_t>(READ_INT_REG<uint64_t>(state, inst->getRs1())));
```
knute-mips, line 65: Hmmm... wonder if this is better:
```suggestion
        const uint64_t rs1_val = READ_INT_REG<uint64_t>(state, inst->getRs1()) & 0xFFFFFFFFull;

```
I assume the `u` in `uw` is unsigned, so that should be ok.  We don't want to sign extend the u64.

DeclanNKim: Yeah, that makes sense. There's quite a few other uw-esque instructions, so I'll go back and pair those down too

core/inst_handlers/b/RvzbaInsts.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+        return ++action_it;
+    }
+
+    template <typename XLEN> 
+    Action::ItrType RvzbaInsts::sh1add_handler(atlas::AtlasState* state, Action::ItrType action_it)
+    {
+        const AtlasInstPtr & inst = state->getCurrentInst();
+
+        const XLEN rs1_val = READ_INT_REG<uint64_t>(state, inst->getRs1());
```
kathlenemagnus, line 79: You should use `XLEN` for the register macro template param instead of hardcoding `uint64_t`. Otherwise there's a cast from `uint64_t` to `uint32_t` for the RV32 version of the handler.

kathlenemagnus:
Is this PR missing the `RVZBB_INST.py` file?

core/inst_handlers/b/RvzbcInsts.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "system/AtlasSystem.hpp"
+
+
+namespace atlas
+{
+    template <typename XLEN>
+    void RvzbcInsts::getInstHandlers(std::map<std::string, Action> & inst_handlers)
+    {
+        static_assert(std::is_same_v<XLEN, RV64> || std::is_same_v<XLEN, RV32>);
+        if constexpr (std::is_same_v<XLEN, RV64>)
```
kathlenemagnus, line 17: To reduce the code size here, you can remove the if-else block and use `XLEN` for the instruction handler template param instead of `RV32` and `RV64`.

DeclanNKim:
This version of the code should have less conditional statements, especially in the ZBB and ZBC instructions. Though I feel like the orc.b instruction could definitely be improved in some way.

core/inst_handlers/b/RvzbbInsts.cpp
klingaard, line 1: A lot of the functions in this file have a pattern, but only deviate by the operation.  For example:
```
    template <typename XLEN> 
    Action::ItrType RvzbbInsts::clzHandler(atlas::AtlasState* state, Action::ItrType action_it)
    {
        const AtlasInstPtr & inst = state->getCurrentInst();

        const XLEN rs1_val = READ_INT_REG<XLEN>(state, inst->getRs1());

        const XLEN rd_val = std::countl_zero(rs1_val);  //  <<<--- the only difference.
        WRITE_INT_REG<XLEN>(state, inst->getRd(), rd_val);

        return ++action_it;
    }
```
Consider creating 1 function that takes the operation as a template parameter:
```
    template <typename XLEN, typename OP> 
    Action::ItrType RvzbbInsts::opHandler(atlas::AtlasState* state, Action::ItrType action_it)
    {
        const AtlasInstPtr & inst = state->getCurrentInst();

        const XLEN rs1_val = READ_INT_REG<XLEN>(state, inst->getRs1());

        const XLEN rd_val = OP()(rs1_val);
        WRITE_INT_REG<XLEN>(state, inst->getRd(), rd_val);

        return ++action_it;
    }
```
Example registration:
```
      inst_handlers.emplace(
                "clz", atlas::Action::createAction<&RvzbbInsts::clzHandler<RV64, std::countl_zero>, RvzbbInsts>(
                            nullptr, "clz", ActionTags::EXECUTE_TAG));
      inst_handlers.emplace(
                "min", atlas::Action::createAction<&RvzbbInsts::minHandler<RV64, std::min>, RvzbbInsts>(
                            nullptr, "min", ActionTags::EXECUTE_TAG));
```

I.e. let the compiler write the code for you.

kathlenemagnus:
@DeclanNKim this is looking great! Just curious - what's the pass rate of the RISC-V arch tests? You can run them with this script from a `release/sim` dir:
```
../../scripts/RunArchTests.py --riscv-arch ../../riscv-tests/isa
```

Also, just wanted to second Knute's comments about template params. I would challenge yourself you remove as much duplicate code as you can! Just so you know, if you need to use an if-statement for a template parameter, it will be compiled out if you use `constexpr` like this:
```
if constexpr (template_param == value)
```

DeclanNKim:
@kathlenemagnus right now, the script tells me that 552/1127 tests passed with the expected rate being 402/558. Most (if not all) of the failed tests were the ones ending with :Zone.Identifier, though I am not quite sure what those mean. Also, I'll continue working towards pairing down as many instructions as I can.

DeclanNKim:
I managed to pair down the ZBA, ZBB, and ZBS instructions. 
 - ZBA: I feel that ZBA can be written in a more succinct way rather than having an unsigned word flag, but I am unsure how.
 - ZBB: I was unable to find a way to efficiently combine the rol and rolw functors (same with ror and rorw) without include some sort of conditional statement, so I left them as two separate functors. I also did not know how to incorporate orc.b into any of these template handlers.
 - ZBC: I did not know how to efficiently further group these 3 instructions into template handlers, so I left them as is.

kathlenemagnus (approved):

PR #165, opened 2025-06-06
Program stack support + a few more syscalls/bug fixes

klingaard:
This work is moving along.  It's still not running a Linux binary to the end, but at least I'm getting to `main`!

I still don't have any testing for this as ELF programs are _still not running_ to the end.  Once I get that working,
then I'll be adding a LOT of testing.

`RunArchTests.py --riscv-arch` results:

```
    PASS RATE: 406/558
EXPECTED RATE: 406/558
```


system/SystemCallEmulator.hpp
```diff
@@ -89,12 +89,6 @@ namespace atlas
 
         std::string workload_;
 
-        //! Bad handler
-        void badSystemCallHandler_(const SystemCallStack &, sparta::memory::BlockingMemoryIF*)
```
kathlenemagnus, line 93: Was this being used anywhere before? How are we handling unimplemented sys calls?

system/AtlasSystem.hpp
```diff
@@ -46,6 +47,9 @@ namespace atlas
         // Get starting PC from ELF
         Addr getStartingPc() const { return starting_pc_; }
 
+        // Get the workload and its program arguments
+        std::vector<std::string> getWorkloadAndArgs() const { return workload_and_args_; }
```
kathlenemagnus, line 51: Can we return a reference instead of a copy?

kathlenemagnus (approved):

core/AtlasState.cpp
```diff
@@ -445,6 +445,161 @@ namespace atlas
...
+
+        // Get the SP (aka x2)
+        const bool MUST_EXIST = true;
+        sparta::Register* reg = findRegister("sp", MUST_EXIST);
+        uint64_t sp = 0;
+        if (xlen_ == 64) {
+            sp = reg->dmiRead<RV64>();
+        }
+        else {
+            sp = reg->dmiRead<RV64>();
```
kathlenemagnus, line 504: Should be `RV32`.

core/AtlasState.cpp
```diff
@@ -445,6 +445,161 @@ namespace atlas
...
+                      "to set it to something...smarter");
+
+        auto* memory = sparta::notNull(atlas_system_)->getSystemMemory();
+        sparta_assert(memory != nullptr,
+                      "Got no memory to preload with the argument stack");
+
+        ////////////////////////////////////////////////////////////////////////////////
+        // Set up argc
+        uint64_t data = program_arguments.size();
+        memory->poke(sp, 8, (uint8_t*)&data);
```
kathlenemagnus, line 517: Is there any difference between RV32 and RV64 here? If so, add a template param. We can assert for RV32 if you're not ready to support it yet.

system/SystemCallEmulator.hpp
klingaard: It was dead code.  I implemented a system call # to a function callback.  If the system call number is not found/implemented, then you get this error message: "System call #123 is not known"

system/AtlasSystem.hpp
klingaard: Yep

core/AtlasState.cpp
klingaard: Added the missing asserts

core/inst_handlers/a/RvaInsts.cpp
```diff
@@ -322,8 +322,10 @@ namespace atlas
         {
             rd_val = state->readMemory<SIZE>(paddr);
         }
-        inst->getRdReg()->write(rd_val);
+        // Must read the RS2 value before writing the Rd (might be the
+        // same register!)
         const RV rs2_val = inst->getRs2Reg()->dmiRead<uint64_t>();
+        inst->getRdReg()->write(rd_val);
```
klingaard, line 328: @kathlenemagnus -- this small line flip was the difference between running a hello world app and not!  :rofl:

klingaard:
Ok, went a _little_ crazy and found two issues: 

1. `amoHandler` was writing RD before ready RS2, which is bad if they are the same register.
2. `exit_group` system call was missing

After fixing these two issues, I got a running "Hello World" linux app!  Dhystone still doen't run.  :frowning:

PR #168, opened 2025-06-11
Testing Out a Post-Merge Action

kathlenemagnus:
First step in adding a regression that is run after every merge to the main branch.


klingaard (approved):
Wonder if you can use artifacts from the merge PR that triggered this

PR #177, opened 2025-06-13
More Fixes to Badge Generation

kathlenemagnus:


colby-nyce (approved):

kathlenemagnus:
Apparently repository secrets are not accessible in pull requests from a fork - so what I have may work once when the workflow is triggered on the push to the main branch. Let's give it a try!

PR #181, opened 2025-06-17
Fixed Several FP Bugs Found in Tenstorrent Tests

kathlenemagnus:
The Tenstorrent tests found a handful of FP bugs. Softfloat uses an awful global variable for the rounding mode, which was not being initialized for some instruction handlers. Also found issues in several "multiply-add" type instructions where 2 softfloat functions (`mul` and `add`) were being called instead of 1 (`mulAdd`). For some rounding modes, this caused the results to be slightly different since softfloat rounds after each function call.


klingaard (approved):
Woo hoo!

core/inst_handlers/d/RvdInsts.cpp
```diff
@@ -183,9 +183,9 @@ namespace atlas
         const uint64_t rs1_val = READ_FP_REG<RV64>(state, inst->getRs1());
         const uint64_t rs2_val = READ_FP_REG<RV64>(state, inst->getRs2());
         const uint64_t rs3_val = READ_FP_REG<RV64>(state, inst->getRs3());
-        const uint64_t product = f64_mul(float64_t{rs1_val}, float64_t{rs2_val}).v;
-        WRITE_FP_REG<RV64>(state, inst->getRd(),
-                           f64_add(float64_t{product ^ (1UL << 63)}, float64_t{rs3_val}).v);
+        static constexpr uint64_t sign_mask = 1UL << 63;
```
colby-nyce, line 186: Prefer non-static constexpr's at the top of the cpp file (anonymous namespace) for fastest runtime.

PR #184, opened 2025-06-19
Added VLEN Parameter and Refactored Observers to Support Vector Registers

kathlenemagnus-mips:
Added support for VLENs of 128, 256, 512, 1024 and 2048. Each VLEN requires a separate vector register JSON. Also refactored the Observer classes to store register values as byte vectors. This is required since vector registers are larger than 64-bits and cannot be stored in a standard type.

### DIFF
--- a/core/AtlasState.cpp
+++ b/core/AtlasState.cpp
@@ -41,6 +41,7 @@ namespace atlas
         sparta::Unit(core_tn),
         hart_id_(p->hart_id),
         isa_string_(p->isa_string),
+        vlen_(p->vlen),
         xlen_(getXlenFromIsaString_(isa_string_)),
         supported_isa_string_(std::string("rv" + std::to_string(xlen_) + "gcbv_zicsr_zifencei")),
         isa_file_path_(p->isa_file_path),
@@ -64,8 +65,9 @@ namespace atlas
         int_rset_ =
             RegisterSet::create(core_tn, json_dir + std::string("/reg_int.json"), "int_regs");
         fp_rset_ = RegisterSet::create(core_tn, json_dir + std::string("/reg_fp.json"), "fp_regs");
+        const std::string vec_reg_json = "/reg_vec" + std::to_string(vlen_) + ".json";
         vec_rset_ =
-            RegisterSet::create(core_tn, json_dir + std::string("/reg_vec.json"), "vec_regs");
+            RegisterSet::create(core_tn, json_dir + vec_reg_json, "vec_regs");
         csr_rset_ =
             RegisterSet::create(core_tn, json_dir + std::string("/reg_csr.json"), "csr_regs");
 
--- a/core/AtlasState.cpp
+++ b/core/AtlasState.cpp
@@ -66,8 +66,7 @@ namespace atlas
             RegisterSet::create(core_tn, json_dir + std::string("/reg_int.json"), "int_regs");
         fp_rset_ = RegisterSet::create(core_tn, json_dir + std::string("/reg_fp.json"), "fp_regs");
         const std::string vec_reg_json = "/reg_vec" + std::to_string(vlen_) + ".json";
-        vec_rset_ =
-            RegisterSet::create(core_tn, json_dir + vec_reg_json, "vec_regs");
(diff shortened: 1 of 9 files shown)


kathlenemagnus-mips:
@colby-nyce where is this "write mask" from `CSRFieldIdxs64.hpp` used?
```
        // Register V0
        struct V0 {
            static constexpr uint32_t reg_num = 0;
            static constexpr uint64_t writable_fields_bit_mask = 0xffffffffffffffff;
        };
```
The problem I have is that vector registers are larger than 64-bits so their writable mask cannot be stored in a `uint64_t`. However, they are always writable so the mask isn't needed anyways. I don't see the mask being used in Atlas today - has their use been deprecated?

core/observers/Observer.hpp
```diff
@@ -56,37 +56,86 @@ namespace atlas
...
+            RegValue(const std::vector<uint8_t> & value) : value_(value) {}
+
+            template <typename SIZE> RegValue(SIZE value) { setValue<SIZE>(value); }
+
+            void setValue(const std::vector<uint8_t> & value)
+            {
+                value_ = value;
+            }
+
+            template <typename SIZE> void setValue(SIZE value)
```
klingaard, line 74: This is a little wonky.  Does this method always expect a POD type for `SIZE`?  I'm wondering if `SIZE` should be called `VAL_TYPE` or something.

core/observers/Observer.hpp
```diff
@@ -177,7 +226,14 @@ namespace atlas
             mem_writes_.clear();
         }
 
-        uint64_t readRegister_(const sparta::Register* reg);
+        std::vector<uint8_t> readRegister_(const sparta::Register* reg)
```
klingaard, line 229: `const`?  :beer:

core/observers/Observer.hpp
kathlenemagnus-mips: Yah I guess `TYPE` would be better. There are probably some asserts I can add to make sure that the type is a standard type like `uint32_t`, `uint64_t`, etc.

klingaard: `static_assert` https://en.cppreference.com/w/cpp/types/is_pod.html

kathlenemagnus-mips: Compiler told me to use `is_standard_layout` and `is_trivial` instead so that's what I did!
```
 error: 'is_pod_v<unsigned long>' is deprecated: use is_standard_layout_v && is_trivial_v instead [-Werror,-Wdeprecated-declarations]
```

colby-nyce:
> @colby-nyce where is this "write mask" from `CSRFieldIdxs64.hpp` used?
> 
> ```
>         // Register V0
>         struct V0 {
>             static constexpr uint32_t reg_num = 0;
>             static constexpr uint64_t writable_fields_bit_mask = 0xffffffffffffffff;
>         };
> ```

This was originally used in the first implementation of the register macros. Basically I used it to pick a more performant call under the hood for some masks (using constexpr / enable_if).

Since nobody uses it anymore though so I can take a note to remove this code from GenCSRHeaders.py

core/observers/Observer.hpp
```diff
@@ -56,37 +56,90 @@ namespace atlas
...
+
+            void setValue(const std::vector<uint8_t> & value)
+            {
+                value_ = value;
+            }
+
+            template <typename TYPE> void setValue(TYPE value)
+            {
+                static_assert(std::is_trivial<TYPE>());
+                static_assert(std::is_standard_layout<TYPE>());
```
colby-nyce, line 77: Should you also `static_assert(std::is_integral_v<TYPE>)` ?

colby-nyce (approved):

core/AtlasState.cpp
```diff
@@ -41,6 +41,7 @@ namespace atlas
         sparta::Unit(core_tn),
         hart_id_(p->hart_id),
         isa_string_(p->isa_string),
+        vlen_(p->vlen),
```
lzhu-pub, line 44: is it also necessary to call `vector_config_->setVLEN()`?

PR #189, opened 2025-06-24
v ext: adding vector mask instruction handlers.

lzhu-pub:
Added all instruction handlers under vector mask category.
Added unit tests for all implemented instruction handlers.


core/inst_handlers/v/RvvmInsts.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include "core/Action.hpp"
+
+namespace atlas
+{
+    class AtlasState;
+    class Action;
+    class ActionGroup;
+
+    class RvvmInsts
```
kathlenemagnus-mips, line 15: It's starting to get hard to decode these class and file names. How's `RvvMaskInsts` instead? Let's update the other classes and file names in the same manner:
```
RvvIntegerInsts
RvvLoadStoreInsts
```

kathlenemagnus-mips (approved):
I appreciate the VERY thorough testing!

kathlenemagnus (approved):

PR #194, opened 2025-07-15
V float: adding vector float and more vector integer instruction handlers

lzhu-pub:
No unit tests available for vector float handlers.


core/VecElements.hpp
```diff
@@ -206,12 +201,52 @@ namespace atlas
          */
         void setVal(ValueType value)
         {
-            ValueType bitmask = getMask();
-            val_ = value & bitmask;
-            ValueType val = READ_VEC_ELEM<ValueType>(state_, reg_id_, idx_) & ~bitmask;
+            ValueType mask = getMask();
```
knute-mips, line 204: Can be `const`

core/VecElements.hpp
```diff
@@ -206,12 +201,52 @@ namespace atlas
...
 
+        /**
+         * @brief Retrieve bit of *Element* at *index* from vector register.
+         * @param index The index of requested bit.
+         * @return Bit value at *index*.
+         */
+        ValueType getBit(size_t index)
+        {
+            val_ = READ_VEC_ELEM<ValueType>(state_, reg_id_, idx_) & getMask();
+            ValueType bitmask = (ValueType{1} << index) & getMask();
```
knute-mips, line 218: Can be `const`

knute-mips, line 218: Why are you masking the vector register element and then masking again for the bit mask?

core/VecElements.hpp
```diff
@@ -206,12 +201,52 @@ namespace atlas
...
+
+        /**
+         * @brief Update bit of *Element* at *index* in vector register to 1.
+         * @param index The index of requested bit.
+         */
+        void setBit(size_t index)
+        {
+            ValueType val = READ_VEC_ELEM<ValueType>(state_, reg_id_, idx_);
+            ValueType mask = getMask();
+            ValueType bitmask = (ValueType{1} << index) & getMask();
```
knute-mips, line 230: You create a local `mask` and then call `getMask()` in the next line.  Make the mask a `const`.

core/VecElements.hpp
```diff
@@ -206,12 +201,52 @@ namespace atlas
...
+
+        /**
+         * @brief Update bit of *Element* at *index* in vector register to 0.
+         * @param index The index of requested bit.
+         */
+        void clearBit(size_t index)
+        {
+            ValueType val = READ_VEC_ELEM<ValueType>(state_, reg_id_, idx_);
+            ValueType mask = getMask();
+            ValueType bitmask = (ValueType{1} << index) & getMask();
```
knute-mips, line 244: Two `getMask` calls.  Need to make `mask` const

core/VecElements.hpp
```diff
@@ -359,13 +396,12 @@ namespace atlas
...
 
-        template <bool M = isMaskElems>
-        requires EnableIf<M>
-
         /**
          * @class MaskBitIterator
          * @brief Used for bit iteration of *MastElements*.
          */
+        template <bool M = isMaskElems>
+        requires EnableIf<M>
```
knute-mips, line 404: I understand the concept you're adding, but I don't understand the need for it in this context.  Do you just not want iteration support for non-mask elements?  I.e. you want that to be a compiler error?

core/VecElements.hpp
```diff
@@ -264,9 +299,11 @@ namespace atlas
      * @tparam ElemType Element type for individual element.
      * @tparam isMaskElems true if this *Elements* represents vector mask.
      */
-    template <typename ElemType, bool isMaskElems> class Elements
+    template <typename EType, bool isMaskElems> class Elements
```
knute-mips, line 302: Curious, can you not determine the boolean `isMaskElems` from the `ElemType`?

core/inst_handlers/finsts_helpers.hpp
```diff
@@ -0,0 +1,295 @@
...
+        WRITE_CSR_REG<XLEN>(state, FCSR, (fcsr & ~mask) | value);
+    }
+
+    template <typename T> struct FuncTraits;
+
+    template <typename R, typename... Args> struct FuncTraits<R (*)(Args...)>
+    {
+        using ReturnType = R;
+        using ArgsTuple = std::tuple<Args...>;
+    };
```
knute-mips, line 58: These are only used by the helper/wrapper function below.  Move these down there...

core/inst_handlers/inst_helpers.hpp
```diff
@@ -22,7 +22,7 @@ template <typename T> inline reg_t zext(T x, uint32_t pos)
 inline uint64_t mulhu(uint64_t a, uint64_t b)
 {
     uint64_t t;
-    uint32_t y1, y2, y3;
+    uint32_t y1, y2;
```
knute-mips, line 25: Variables uninitialized.

core/inst_handlers/v/RvvFloatInsts.hpp
```diff
@@ -0,0 +1,81 @@
...
+        {
+            enum struct Mode
+            {
+                V, // vector
+                F, // scaler
+                I, // immediate
+                W  // wide
+            };
+
+            Mode dst, src1, src2;
```
knute-mips, line 37: These should be `const`

core/inst_handlers/v/RvvFloatInsts.hpp
```diff
@@ -0,0 +1,81 @@
...
+            constexpr OperandMode(Mode dst, Mode src2, Mode src1) : dst(dst), src1(src1), src2(src2)
+            {
+            }
+        };
+
+        template <typename F16, typename F32, typename F64> struct FloatFuncs
+        {
+            F16 f16;
+            F32 f32;
+            F64 f64;
```
knute-mips, line 48: These can be `const`

core/inst_handlers/v/RvvIntegerInsts.hpp
```diff
@@ -15,15 +15,44 @@ namespace atlas
...
+        {
+            enum struct Mode
+            {
+                V, // vector
+                X, // scaler
+                I, // immediate
+                W  // wide
+            };
+
+            Mode dst, src1, src2;
```
knute-mips, line 28: `const`

core/inst_handlers/v/RvvFloatInsts.cpp
```diff
@@ -0,0 +1,1287 @@
...
+        Elements<
+            Element<opMode.dst == RvvFloatInsts::OperandMode::Mode::W ? 2 * elemWidth : elemWidth>,
+            false>
+            elems_vd{state, state->getVectorConfig(), inst->getRd()};
+        softfloat_roundingMode = READ_CSR_REG<XLEN>(state, FRM);
+
+        auto execute = [&]<typename Iterator>(const Iterator & begin, const Iterator & end)
+        {
+            for (auto iter = begin; iter != end; ++iter)
+            {
```
knute-mips, line 904: Why bother with a copy of begin?
```suggestion
        auto execute = [&](auto iter, const auto & end)
        {
            for (;iter != end; ++iter)
            {
```

knute-mips (commented):
This is a big PR, so I'm not done, but I'll submit what I've found so far.

core/VecElements.hpp
```diff
@@ -206,12 +201,52 @@ namespace atlas
...
+            ValueType val = READ_VEC_ELEM<ValueType>(state_, reg_id_, idx_) & ~mask;
             WRITE_VEC_ELEM<ValueType>(state_, reg_id_, val | val_, idx_);
         }
 
+        /**
+         * @brief Retrieve bit of *Element* at *index* from vector register.
+         * @param index The index of requested bit.
+         * @return Bit value at *index*.
+         */
+        ValueType getBit(size_t index)
```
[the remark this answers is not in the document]
lzhu-pub: no. `val_` is member variable and it's updated from REG value.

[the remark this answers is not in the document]
knute-mips: Suggest a name change: `getBit` -> `updateBit`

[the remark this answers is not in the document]
lzhu-pub: That name seems slightly off, since the bit in `val_` is not updated. The purpose of the function is to get the bit from `val_`, which is "by-the-way" up-to-date.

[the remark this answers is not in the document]
knute-mips: Side effects of functions that appear to be only read-only is hazardous.  What if I call this method back to back to back?   I should get the same answer, which if you look at the function you do.  So what's the point of updating a private variable?  Suggest making this a proper read-only method.

core/VecElements.hpp
lzhu-pub: I agree this is the best solution. it also makes the `get` read-only.

core/VecElements.hpp
```diff
@@ -206,12 +201,52 @@ namespace atlas
...
+            ValueType val = READ_VEC_ELEM<ValueType>(state_, reg_id_, idx_) & ~mask;
             WRITE_VEC_ELEM<ValueType>(state_, reg_id_, val | val_, idx_);
         }
 
+        /**
+         * @brief Retrieve bit of *Element* at *index* from vector register.
+         * @param index The index of requested bit.
+         * @return Bit value at *index*.
+         */
+        ValueType getBit(size_t index) const
```
kathlenemagnus, line 215: These are good functions, but we should considering moving some of this logic into a utils header file. I'm sure there are other places in the code where these functions could be used.

core/translate/AtlasTranslationState.hpp
```diff
@@ -9,7 +9,7 @@ namespace atlas
     class AtlasTranslationState
     {
       public:
-        static const uint32_t MAX_TRANSLATION = 8;
+        static const uint32_t MAX_TRANSLATION = 64;
```
kathlenemagnus, line 12: Can you calculate this from other consts? Like `MAX_VLEN/MIN_SEW` or something like that.

kathlenemagnus:
WOW! Lots of good stuff in this PR. Looking forward to being able to test it!

core/VecElements.hpp
lzhu-pub: Since the mask is specific to VectorElement, this version of `getBit()` can only be used on vector structures. More generic `getBit()` will need a different mask handling.

core/inst_handlers/v/RvvIntegerInsts.cpp
```diff
@@ -62,42 +621,327 @@ namespace atlas
...
+    template <typename XLEN, RvvIntegerInsts::OperandMode opMode, bool hasMaskOp, auto detectFunc>
+    Action::ItrType RvvIntegerInsts::vmiacsbHandler_(AtlasState* state, Action::ItrType action_it)
     {
-        (void)state;
+        VectorConfig* vector_config = state->getVectorConfig();
+        switch (vector_config->getSEW())
+        {
+            case 8:
+                return vmiacsbHelper<XLEN, 8, opMode, hasMaskOp, detectFunc>(state, action_it);
+                break;
```
colby-nyce, line 831: None of these break statements are necessary.

PR #196, opened 2025-07-18
Vector reduction sum instruction handler

skhanx86:
This PR contains the instruction handler support for reduction sum instruction. (vredsum.vs)


core/inst_handlers/v/RvvReductionInsts.cpp
```diff
@@ -0,0 +1,122 @@
...
+                nullptr,
+                "vredsum.vs",
+                ActionTags::EXECUTE_TAG));
+    }
+
+    // Template instantiations for both RV32 and RV64
+    template void RvvReductionInsts::getInstHandlers<RV32>(std::map<std::string, Action>&);
+    template void RvvReductionInsts::getInstHandlers<RV64>(std::map<std::string, Action>&);
+
+    // Helper function for actual reduction logic
```
lzhu-pub, line 28: i believe eventually you will choose lambda.

test/sim/vro/Vred_test.cpp
```diff
@@ -99,7 +99,10 @@ class VredInstructionTester : public PegasusInstructionTester
         offset += 7;
         opcode |= rd << offset; // rd (scalar result)
         offset += 5;
-        opcode |= 0 << offset; // funct3 = 000
+        if(isWideningEnabled)
```
kathlenemagnus, line 102: Use brackets on if-else blocks.

core/inst_handlers/v/RvvReductionInsts.cpp
```diff
@@ -0,0 +1,85 @@
...
+    Elements<Element<ElemWidth>, false> elems_vs1{state, state->getVectorConfig(), inst->getRs1()};
+    Elements<Element<ElemWidth>, false> elems_vd{state, state->getVectorConfig(), inst->getRd()};
+
+    auto accumulator = elems_vs1.getElement(0).getVal();
+
+    auto execute = [&]<typename Iterator>(Iterator begin, Iterator end)
+    {
+        for (auto iter = begin; iter != end; ++iter)
+        {
+                const auto idx = iter.getIndex();
```
kathlenemagnus, line 42: Spacing is weird - are there tabs here?

core/inst_handlers/v/RvvReductionInsts.cpp
```diff
@@ -0,0 +1,85 @@
...
+    {
+        execute(elems_vs2.begin(), elems_vs2.end());
+    }
+    else // masked
+    {
+        const MaskElements mask_elems{state, state->getVectorConfig(), pegasus::V0};
+        execute(mask_elems.maskBitIterBegin(), mask_elems.maskBitIterEnd());
+    }
+
+    elems_vd.getElement(0).setVal(accumulator);
```
kathlenemagnus, line 57: What does this instruction write to the other elements of VD?

core/inst_handlers/v/RvvReductionInsts.cpp
```diff
@@ -0,0 +1,85 @@
...
+    return ++action_it;
+}
+
+
+
+    // Dispatch SEW-sized implementation of vredsum.vs
+    template <template <typename> typename OP>
```
kathlenemagnus, line 64: I feel like there should be a more simple way to define these template params. Have you tried just `template<typename OPERATOR>`?

kathlenemagnus (approved):

core/inst_handlers/v/RvvReductionInsts.cpp
skhanx86: Yes removed the non lambda, commented version.

core/inst_handlers/v/RvvReductionInsts.cpp
skhanx86: I considered template<typename OPERATOR>, but in this context I need a template that can be instantiated per SEW type (e.g. std::plus<uint8_t>, std::plus<uint16_t>, etc.). Using a template template parameter allows that. If I used template<typename OPERATOR>, I’d lose the ability to specialize per element width in the handler function.

core/inst_handlers/v/RvvReductionInsts.cpp
skhanx86: I used tabs, wouldn't a make clangformat fix/manage this?

kathlenemagnus: Yes running `make clangformat` should fix this. If it doesn't, please manually fix this before merging.

core/inst_handlers/v/RvvReductionInsts.cpp
kathlenemagnus: Can you put a TODO here for when we add a parameter to specify the agnostic policy? It's common for the agnostic policy to actually write all 1's to the tail since undisturbed requires adding the dest register as a source.

core/inst_handlers/v/RvvReductionInsts.cpp
lzhu-pub: There are similar issues that has been bugging me for a while. After looking at the update diffs, I found even for untouched files, clangformat would change it based on the other changed files. I presume something like "the average line length to decide the maximum line length for all files"...
This created a lot of unnecessary diffs in PR.

PR #199, opened 2025-07-24
CoSim pipeline (event retrieval only)

colby-nyce:
This PR starts the implementation of the CoSim flush / event retrieval pipeline. Flushing will be added in the next PR, mostly needing to change the `CoSimPipeline::createPipeline()` method and the `cosim::Event` class to support it.


cosim/Event.hpp
```diff
@@ -213,19 +183,36 @@ namespace pegasus::cosim
...
         const std::vector<MemWriteAccess> & getMemoryWrites() const { return memory_writes_; }
 
-        const std::string getDisassemblyStr() const { return mavis_opcode_info_->dasmString(); }
+        const std::string & getDisassemblyStr() const { return dasm_string_; }
 
-        //! @}
-        ////////////////////////////////////////////////////////////////////////////////////////////
+        bool operator==(const Event & other) const
+        {
+            return event_uid_.getValue() == other.event_uid_.getValue() && type_ == other.type_
```
kathlenemagnus, line 190: We should only need to check the UID right? Since it should be uniquely assigned.

cosim/Event.hpp
```diff
@@ -52,6 +52,17 @@ namespace pegasus::cosim
                 value(convertToByteVector(val))
             {
             }
+
+            RegReadAccess() = default;
+
+            /// Called to/from char buffer (boost::serialization)
+            template <typename Archive> void serialize(Archive & ar, const unsigned int /*version*/)
+            {
+                ar & reg_id;
```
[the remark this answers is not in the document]
colby-nyce: The way boost::serialization works is that you write this one method once, and it is called by boost serializers (read from member vars to a byte vector) and deserializers (populate member vars from a byte vector). The two different Archive classes implement operator& differently so we don't need more than one serialize() method.

cosim/Event.hpp
```diff
@@ -294,12 +281,74 @@ namespace pegasus::cosim
         //! \name Mavis Opcode Information
         //! @{
 
-        mavis::OpcodeInfo::PtrType mavis_opcode_info_;
+        std::string dasm_string_;
```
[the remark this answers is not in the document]
colby-nyce: Nothing can be const anymore since `Events` will be in `simdb::ConcurrentQueue<Event>` and there is only one way to get an Event from it:

```
Event evt;
if (queue.try_pop(evt)) { ... }
```

So we need a default ctor and therefore no const members in this case.

test/cosim/CMakeLists.txt
```diff
@@ -1 +1,18 @@
-add_subdirectory(cosim_step)
+macro(cosim_named_test name target)
+    pegasus_named_test(${name} ${target} ${ARGN})
+
+    target_link_libraries(${target} PRIVATE
+        pegasuscosimlib
```
kathlenemagnus, line 5: You should only have to link against `pegasuscosimlib` since that library is already linked with everything below.

test/cosim/cosim_workload/run_workload/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+project(RunWorkload_Test)
+
+add_executable(RunWorkload_test RunWorkload_test.cpp)
+
+file (CREATE_LINK ${CMAKE_SOURCE_DIR}/arch       ${CMAKE_CURRENT_BINARY_DIR}/arch SYMBOLIC)
+file (CREATE_LINK ${CMAKE_SOURCE_DIR}/mavis/json ${CMAKE_CURRENT_BINARY_DIR}/mavis_json SYMBOLIC)
+file (CREATE_LINK ${CMAKE_SOURCE_DIR}/core/rv64  ${CMAKE_CURRENT_BINARY_DIR}/rv64 SYMBOLIC)
+
+cosim_named_test(RunWorkload_test_run RunWorkload_test /home/cnyce/atlas-fork/riscv-tests/isa/rv64mi-p-csr)
```
kathlenemagnus, line 9: Fix path to test. You can copy `rv64mi-p-csr` and stick it in the test directory if you'd like.

cosim/PegasusCoSim.cpp
```diff
@@ -46,99 +45,122 @@ namespace pegasus
...
 
-        const uint32_t num_harts = 1;
         for (uint32_t hart_id = 0; hart_id < num_harts; ++hart_id)
         {
             // Get Fetch for each hart
             const std::string core_name = "core" + std::to_string(hart_id);
             fetch_.emplace_back(
                 getRoot()->getChild(core_name + ".fetch")->getResourceAs<pegasus::Fetch>());
 
+            auto state = getPegasusState(hart_id);
```
kathlenemagnus, line 83: This doesn't need to be fixed in this PR, but we should review what methods are public in `PegasusCoSim`. We will eventually have users (e.g. a performance model, DV environment) that we will need to prevent from doing dangerous things like calling `PegasusState` methods directly.

kathlenemagnus (approved):
Looks great! Thank you for including good comments in the `CoSimPipeline` class. Made it very easy to follow.

PR #202, opened 2025-07-28
Remove -DBUILD_COSIM and bump Sparta ref

colby-nyce:
This workaround can be removed now that all the latest from SimDB for cosim is in Sparta master.

I think the map_v2.1 branch has to be changed first before this gets merged into Pegasus:
- Protect the map_v2.1 branch so it only allows cherry picked commits (as a long-lived release branch off master)
- Update the Pegasus README to mention map_v2.1 (not map_v2)


.github/workflows/ubuntu-build.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           repository: sparcians/map
           submodules: recursive
           path: map
-          ref: map_v2.0.22
+          ref: map_v2.1.0
```
knute-mips, line 41: yes

kathlenemagnus (approved):

PR #203, opened 2025-07-29
V tenstorrent

lzhu-pub:


core/inst_handlers/inst_helpers.hpp
```diff
@@ -66,6 +68,92 @@ inline int64_t mulh(int64_t a, int64_t b)
     return negate ? ~res + ((uint64_t)a * (uint64_t)b == 0) : res;
 }
 
+template <typename T> struct Mulhu
+{
+    inline T operator()(const T & lhs, const T & rhs) const
+    {
+        if constexpr (sizeof(T) < sizeof(uint64_t))
```
kathlenemagnus, line 75: I believe this duplicates code in `RvmInsts::mulhu_64Handler_` which I'm sure is the case for some of these other handlers. too... This is very messy. We should clean up the `RvmInsts` class to use `Mulhu`. We can templatize the instruction handlers in `RvmInsts` and reduce a lot of code duplication in that class as well.

core/inst_handlers/inst_helpers.hpp
```diff
@@ -66,6 +68,92 @@ inline int64_t mulh(int64_t a, int64_t b)
...
+{
+    inline const T & operator()(const T & lhs, const T & rhs) const { return std::max(lhs, rhs); }
+};
+
+template <typename T> struct Min
+{
+    inline const T & operator()(const T & lhs, const T & rhs) const { return std::min(lhs, rhs); }
+};
+
+template <typename T> struct Sll
```
kathlenemagnus, line 132: Again, code duplication in `RviInsts`. Any shared functionality between the vector extension and other extensions needs to be defined in a common place. If we have multiple definitions of these common operators it's going to be a nightmare to debug.

lzhu-pub: As for the location of inst_helerps.hpp, it's already in a common place.
I agree we should clean up to use template functors, in case we want to eliminate duplications.
But the point here is, these new codes are template functors. It's only used where only a generic  functor hammer is needed to fulfill the job. The same logic is behind `*`  vs `std::multiples`. Do we want to use `std::multiplies` in all places?

core/inst_handlers/inst_helpers.hpp
lzhu-pub: At first I look at the original `mulhu()`, I was wondering why we didn't have it templated. Then I convinced myself that we have it only for uint64_t is because, maybe, we can just extend the other shorter types to uint64_t and extract high bits. 
I am trying to reuse `mulhu(uint64_t, uint64_t)` here as well.

kathlenemagnus (approved):

PR #207, opened 2025-07-31
Added Zcmp Extension Support

kathlenemagnus:
This PR also includes:
- Instruction limit command line parameter to end simulation when a specified instruction count is reached
- Opcode command line parameter to execute a single opcode (useful for debugging)
- Observers now capture all source and destination registers instead of just RS1, RS2, RS3, RD and RD2.
- Updates Mavis


core/inst_handlers/zcmp/RvzcmpInsts.cpp
```diff
@@ -0,0 +1,159 @@
...
+    template void RvzcmpInsts::getInstHandlers<RV32>(Execute::InstHandlersMap &);
+    template void RvzcmpInsts::getInstHandlers<RV64>(Execute::InstHandlersMap &);
+
+    template <typename XLEN> void RvzcmpInsts::pop_(pegasus::PegasusState* state)
+    {
+        const PegasusInstPtr & inst = state->getCurrentInst();
+
+        // Load RA (1) and 0-12 saved registers (8-9, 18-27) from the stack frame
+        // Have to start at idx 1 since Mavis includes the SP in the list of dests
+        XLEN addr = READ_INT_REG<XLEN>(state, 2);
```
klingaard, line 45: Isn't there an constexpr for 2?

core/inst_handlers/zcmp/RvzcmpInsts.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    template <typename XLEN>
+    Action::ItrType RvzcmpInsts::mvsa01Handler_(pegasus::PegasusState* state,
+                                                Action::ItrType action_it)
+    {
+        const PegasusInstPtr & inst = state->getCurrentInst();
+        const XLEN a0_val = READ_INT_REG<XLEN>(state, 10);
+        const XLEN a1_val = READ_INT_REG<XLEN>(state, 11);
+
+        // FIXME: Does Mavis handle this?
```
klingaard, line 146: I wondered the same thing.  I'm in the middle of fix this instruction right now.  I can make that illegal

core/PegasusInst.hpp
```diff
@@ -118,6 +118,11 @@ namespace pegasus
             return opcode_info_->getSpecialField(mavis::OpcodeInfo::SpecialField::VM);
         }
 
+        uint64_t getStackAdjustment() const
+        {
+            return opcode_info_->getSpecialField(mavis::OpcodeInfo::SpecialField::STACK_ADJ);
```
klingaard, line 123: The stack adjustment can be negative.  I ran into this issue with our internal model.  Does the code handle this correctly?

core/PegasusState.cpp
```diff
@@ -457,6 +493,17 @@ namespace pegasus
         // Increment instruction count
         ++sim_state_.inst_count;
 
+        if constexpr (CHECK_ILIMIT)
+        {
+            if (sim_state_.inst_count == ilimit_)
```
klingaard, line 498: 🥳

sim/pegasus.cpp
```diff
@@ -59,6 +61,7 @@ int main(int argc, char** argv)
             ("inst-limit,i", po::value<uint64_t>(&ilimit),
              "Stop simulation after the instruction limit has been reached")
             ("reg", po::value<std::vector<RegOverride>>()->multitoken(), "Override initial value of a register")
+            ("opcode", po::value<std::string>(&opcode), "Executes a single opcode")
```
klingaard, line 64: 🤤

klingaard (approved):
Only concern I have is with the sign on the stack adjustment.  It might need to be a signed type...

core/PegasusInst.hpp
kathlenemagnus: It seems like it does! The opcode I was testing with had a -32 stack adjustment and I got the expected result. The `getSpecialField` method from Mavis only returns unsigned.

core/inst_handlers/zcmp/RvzcmpInsts.cpp
```diff
@@ -0,0 +1,159 @@
...
+        const PegasusInstPtr & inst = state->getCurrentInst();
+
+        // Load RA (1) and 0-12 saved registers (8-9, 18-27) from the stack frame
+        // Have to start at idx 1 since Mavis includes the SP in the list of dests
+        XLEN addr = READ_INT_REG<XLEN>(state, 2);
+        const XLEN new_sp_val = addr + inst->getStackAdjustment();
+        const auto & dst_reg_list = inst->getMavisOpcodeInfo()->getDestOpInfoList();
+        for (uint32_t idx = 1; idx < dst_reg_list.size(); ++idx)
+        {
+            const auto dst = dst_reg_list.at(idx);
```
colby-nyce, line 50: Might as well just use operator[] since you're in the dst_reg_list loop

core/inst_handlers/zcmp/RvzcmpInsts.cpp
kathlenemagnus: Yes, I replaced the 2 with `SP`.

core/inst_handlers/zcmp/RvzcmpInsts.cpp
kathlenemagnus: Good point! I'll make that change.

PR #208, opened 2025-08-01
change to use float functors in vf f and d extensions

lzhu-pub:


kathlenemagnus (approved):
Looking good! Great clean up 👍

colby-nyce (approved):

PR #210, opened 2025-08-01
Adding memory accesses and change of flow events to a STF Trace

ayeshanh:
This PR records memory reads/writes and change of flow events to a STF Trace. I also updated the initialization of a STFLogger object to accommodate if the user writes the filename with or without the stf file extension.


core/observers/STFLogger.cpp
```diff
@@ -8,8 +8,21 @@ namespace pegasus
                          PegasusState* state) :
         Observer((reg_width == 32) ? ObserverMode::RV32 : ObserverMode::RV64)
     {
-        // set up version and stf generation type
-        stf_writer_.open(filename);
+        if (filename.find(".stf", filename.length() - 5)
```
kathlenemagnus, line 11: Instead of using the `find` method, I found a file system method that returns just the file extension:
https://en.cppreference.com/w/cpp/filesystem/path/extension.html

If you prefer to use `find`, I would make a const list of allowed extensions and use the `find_if` algorithm:
https://en.cppreference.com/w/cpp/algorithm/find.html

core/observers/STFLogger.cpp
```diff
@@ -23,76 +36,90 @@ namespace pegasus
...
+                                    stf::Registers::STF_REG_TYPE::CSR,
+                                    stf::Registers::STF_REG_OPERAND_TYPE::REG_DEST,
+                                    READ_CSR_REG<uint64_t>(state, dst_reg.reg_id.reg_num));
+                                    break;
                 default:
                     sparta_assert(false, "Invalid register type!");
             }
         }
 
+        if (fault_cause_.isValid() || interrupt_cause_.isValid()) { return; } // TODO: Add support for exceptions
```
kathlenemagnus, line 121: Some faults will have valid opcodes. I would try adding a method to the Exception unit that can tell us which fault codes have valid opcodes. You can get a pointer to the Exception unit through PegasusState.
```
Exception* getExceptionUnit() const { return exception_unit_; }
```

core/observers/STFLogger.cpp
```diff
@@ -8,8 +8,21 @@ namespace pegasus
                          PegasusState* state) :
         Observer((reg_width == 32) ? ObserverMode::RV32 : ObserverMode::RV64)
     {
-        // set up version and stf generation type
-        stf_writer_.open(filename);
+        if (filename.find(".stf", filename.length() - 5)
+            || filename.find(".zstf", filename.length() - 6)
+            || filename.find(".STF", filename.length() - 5)
+            || filename.find(".ZSTF", filename.length() - 6))
```
bdutro, line 14: The STF library is case-sensitive, so `.STF` and `.ZSTF` aren't valid extensions.

core/observers/STFLogger.cpp
```diff
@@ -8,8 +8,21 @@ namespace pegasus
...
+        {
+            std::cout << "STF Trace filename inputted does not end with .stf or .zstf, appending "
+                         ".zstf to the filename."
+                      << std::endl;
+            stf_writer_.open(filename + ".zstf");
+        }
+        else
+        {
+            stf_writer_.open(filename);
```
knute-mips, line 23: @bdutro can confirm or deny, but I think the writer will throw if the file name does not contain any extension that is usable.  If so, I'd wrap the `open` in a `try {} catch {}` and force the user to supply a valid stf filename.

bdutro: That's correct.

core/observers/STFLogger.cpp
```diff
@@ -23,76 +36,90 @@ namespace pegasus
...
 
-    // METHODS
     void STFLogger::postExecute_(PegasusState* state)
     {
-        if (fault_cause_.isValid() || interrupt_cause_.isValid())
+        if (state->getNextPc() != state->getPrevPc() + state->getCurrentInst()->getOpcodeSize())
         {
-            return;
+            stf_writer_ << stf::InstPCTargetRecord(state->getNextPc());
+        }
```
bdutro, line 53: You should only use this record type for taken branches. Right now, this will also insert an `InstPCTargetRecord` for faults, interrupts, syscalls, etc.

bdutro: However, until those other cases are wired up you can replace this with a `ForcePCRecord` for all non-branch cases.

core/observers/STFLogger.cpp
ayeshanh: Since the observer class records the faults and interrupts happening in Pegasus, I put the function you were mentioning in postExecute_ then used fault_cause_ to find which faults are passing through and if they are tied to any valid opcodes.

core/observers/STFLogger.cpp
```diff
@@ -93,6 +112,21 @@ namespace pegasus
             }
         }
 
+        if (fault_cause_.isValid()
+            || interrupt_cause_.isValid()) // TODO: Add support for exceptions
+        {
+            if (fault_cause_.getValue() == FaultCause::INST_ACCESS
```
kathlenemagnus, line 118: There are a few other fault causes that won't have valid opcodes: INST_PAGE_FAULT and INST_ADDR_MISALIGNED. These faults will fire during the PC address translation, so an opcode will never be read from memory. Also, some ILLEGAL_INST faults may have a valid opcode, but what you have is ok for now.

ayeshanh: I'll make sure to implement this in my next PR today.

PR #213, opened 2025-08-05
Fixed Zcmp Bug and Regression Failures

kathlenemagnus:
Fixed a minor bug in one of the Zcmp instruction handlers. Also fixed the post-merge regression by changing the way that compression is enabled. The Tenstorrent vector tests are now also include in the regression.


colby-nyce (approved):

kathlenemagnus:
The regression failures were caused by me changing the default Pegasus ISA string to use the compression sub-extensions (i.e. "zca", "zcd", "zcd") instead of just "c".

It's unclear to me how these sub-extensions should interact with the MISA register. Should enabling "zca" set the "c" field? Or do all of the sub-extensions need to be enabled? What about when the "c" field is set in the MISA register; should all extensions be enabled or just "zca"? I'll open a ticket to address this in the future, but this PR just includes the quick fix of changing the ISA string back.

PR #238, opened 2025-08-27
Added a parameter to stop the simulator on unexpected faults

klingaard:
If the user isn't expecting an ILLOP exception, the parameter `top.core?.exception.params.unexpected_faults` can
be used (as a mask) to stop the simulator.  In the above example:

 `-p top.core?.exception.params.unexpected_faults 0x000004`


kathlenemagnus:
Very handy! Thanks for adding this.

PR #247, opened 2025-09-03
Use ExtensionManager to Track Which Extensions are Enabled

kathlenemagnus-mips:
Cleaned up how Pegasus enables and disables extensions. This fixes the failing `ma_fetch` RISC-V arch tests in the post-merge regression.


core/inst_handlers/zicsr/RvzicsrInsts.cpp
```diff
@@ -430,25 +430,40 @@ namespace pegasus
...
         {
-            const bool f_ext_enabled = READ_CSR_FIELD<XLEN>(state, MISA, "f") == 0x1;
-            const bool d_ext_enabled = READ_CSR_FIELD<XLEN>(state, MISA, "d") == 0x1;
-            if (f_ext_enabled)
+            if (READ_CSR_FIELD<XLEN>(state, MISA, "f") == 0x1)
             {
-                inclusions.emplace("f");
+                ext_manager.enableExtension("f");
+                constexpr bool is_rv32 = std::is_same_v<XLEN, RV32>;
+                if (is_rv32 && state->isCompressionEnabled())
```
kathlenemagnus-mips, line 447: @bdutro I need a way to automatically enable and disable dependent extensions. For example, if I disable "f", then I would expect "zcf" and "zve32f" to also be disabled. Conversely, if I enable "f" and "v" is also enabled, I would expect "zve32f" to also be enabled.

If that doesn't make sense, I think if I was able to get a list of dependent instructions then I could enable the dependent extensions myself. The problem is that Pegasus doesn't understand the relationship between these extensions and I don't have a way to access that information today.

core/PegasusState.cpp
```diff
@@ -253,9 +246,15 @@ namespace pegasus
 
     void PegasusState::changeMavisContext()
     {
-        const mavis::MatchSet<mavis::Pattern> inclusions{inclusions_};
+        std::vector<std::string> enabled_exts;
```
kathlenemagnus-mips, line 249: @bdutro can the `ExtensionManager` handling changing the Mavis context for me? Or would you recommend I just create a new instance of Mavis? This works today, but it's inefficient to have to create the list of Mavis inclusions.

core/inst_handlers/zicsr/RvzicsrInsts.cpp
bdutro-mips: If we have an ISA string like `rv64imafd` and you disable `f`, then `d` will also need to be disabled. If you then re-enable `f`, do you expect `d` to be re-enabled?

bdutro-mips: Nevermind, I have a solution where it will automatically disable/re-enable `d` unless `d` is also explicitly disabled.

core/PegasusState.hpp
```diff
@@ -101,6 +101,11 @@ namespace pegasus
             return extension_manager_;
         }
 
+        bool isCompressionEnabled() const
+        {
+            return extension_manager_.isEnabled("zca");
```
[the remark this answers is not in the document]
kathlenemagnus: So enabling `f/d` when `c` is enabled will implicitly enable `zcf/zcd`, respectively. `zca` is the sub-extension of `c` that does not include any FP so enabling `zca` and `f/d` would not enable `zcf/zcd`. Of course `f` is also required for `d`, but does `zcd` require `zcf`? Not sure on that one...

kathlenemagnus:
This PR is ready to go now. Brett made some improvements to the Mavis ExtensionManager which have been integrated into Pegasus. Overall, the way that Pegasus manages Mavis is much cleaner so thank you @bdutro!

knute-mips (approved):

klingaard (approved):

PR #249, opened 2025-09-04
STF tracing with Vector support

skhanx86:


core/observers/STFLogger.cpp
```diff
@@ -93,36 +100,41 @@ namespace pegasus
             }
             else
             {
-                sparta_assert(false, "STF trace generation does not support vector yet!");
+                stf_writer_ << stf::InstRegRecord(src_reg.reg_id.reg_num, stf_reg_type,
+                                                  stf::Registers::STF_REG_OPERAND_TYPE::REG_SOURCE,
+                                                  readVectorRegister_(state, src_reg.reg_id));
```
kathlenemagnus, line 105: For source registers, you should be getting the register value captured in `src_reg` instead of reading the register directly. You may need to extend the `ObservedReg` class to be able to get the value as a `std::vector<uint64_t>`.

lzhu-pub (approved):
Please address above comments before merging.

core/Exception.cpp
```diff
@@ -53,7 +54,8 @@ namespace pegasus
                                                 : static_cast<XLEN>(fault_cause_.getValue());
         DLOG("Exception code: " << excp_code);
 
-        if (false == is_interrupt) {
+        if (false == is_interrupt)
```
colby-nyce, line 57: This would read better IMO as `if (is_interrupt == false)` or `if (!is_interrupt)`

The way it's written makes me pause and reread it...

kathlenemagnus (approved):

colby-nyce (approved):

colby-nyce:
> To reduce clang-format diffs, the approach I use is to only check-in the files that I touched. I am all ears for other suggestions. But PR nowadays contains too much clang-format diffs, which is not relevant and distracting.

Honestly if this issue is occurring over and over, it is surely due to the fact that the clang-format github check is not currently failing the check even when the check does not pass (i.e. a PR was submitted that did NOT run clang-format, and the next PR has all the diffs)

PR #250, opened 2025-09-05
Fixed Typo in Zfa Instruction Handler Name

kathlenemagnus:


klingaard (approved):
Isn't this fix also in #247?

PR #257, opened 2025-09-15
STFValidator enhancements

skhanx86:
Added OPCODE comparisons and register write mismatch identification for vector registers.


core/observers/STFValidator.cpp
```diff
@@ -73,6 +83,17 @@ namespace pegasus
             sparta_assert(false, "PCs have diverged!");
         }
 
+        if (state->getSimState()->current_opcode != next_it_->opcode() && !(next_it_->isSyscall()))
+        { // Since syscalls are not in Pegasus, can ignore them.
```
kathlenemagnus, line 87: Does this mean that the`STFValidator` is not called when Pegasus executes a syscall? What about when syscall emulation is disabled?

core/observers/STFValidator.cpp
```diff
@@ -51,6 +51,16 @@ namespace pegasus
         }
     }
 
+    std::string formatVectorHex(const std::vector<uint64_t> & vec)
```
kathlenemagnus, line 54: Can you make this a private method of the base `Observer` class? We can use it for the instruction logger, too.

kathlenemagnus (approved):

core/observers/STFValidator.cpp
skhanx86: If I do not skip the syscall, this is what I see from the stfValidator output:

```
{0000000000 00000000 top.core0 stf_valid} postExecute_: OPCODEs do not match!
{0000000000 00000000 top.core0 stf_valid} postExecute_: uid: 206 ecall   [pre execute (<no tag>)] -> [ecall (EXECUTE)] -> finish_inst
{0000000000 00000000 top.core0 stf_valid} postExecute_:     PC: 0x80012238
{0000000000 00000000 top.core0 stf_valid} postExecute_: Pegasus Opcode: 0x115
{0000000000 00000000 top.core0 stf_valid} postExecute_:     STF Opcode: 0x0
```

core/observers/STFValidator.cpp
```diff
@@ -111,12 +122,31 @@ namespace pegasus
...
+                            STFVALIDLOG("0x" << std::hex << pc << std::dec << ":\t"
+                                             << state->getCurrentInst());
                             STFVALIDLOG("    Pegasus value: 0x" << std::hex << reg_val);
                             STFVALIDLOG("        STF value: 0x" << std::hex << stf_reg_val);
                             STFVALIDLOG("");
                         }
                     }
+                    else
+                    {
+                        std::vector<uint64_t> reg_val = readVectorRegister_(state, dst_reg.reg_id);
```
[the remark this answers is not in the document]
skhanx86: Currently, getVectorValue() appears to return only the logical vd, not the full physical group when LMUL > 1. Similarly, we're only reading and validating vd from the simulator.

So unless the trace includes the full group, comparing just vd is appropriate for now. We should revisit this if we start supporting full group writes or multi-reg comparisons in the trace.

core/observers/STFValidator.cpp
skhanx86: Sounds good.

PR #258, opened 2025-09-18
V ext: adding some other unsupported handlers

lzhu-pub:
1. added bunch of unimplemented handlers.
2. added float point reciprocal support.


core/inst_handlers/v/RvvLoadStoreInsts.cpp
```diff
@@ -654,4 +778,37 @@ namespace pegasus
...
         return ++action_it;
     }
+
+    template <bool isLoad>
+    Action::ItrType RvvLoadStoreInsts::vlsmHandler_(pegasus::PegasusState* state,
+                                                    Action::ItrType action_it)
+    {
+        constexpr size_t BYTESIZE = 8;
+        const PegasusInstPtr & inst = state->getCurrentInst();
+        VectorConfig config{*state->getVectorConfig()};
```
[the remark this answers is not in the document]
lzhu-pub: I totally understand your point. There is some detail that we need to iron out before we make the changes.
1. vector `Elements` requires a  pointer of `VecConfig` for construction. We can use either global, or instruction local as you suggested.
2. If we want `VecConfig` to be carried by `PegasusInst`, when do we set its value? During decoding? Theoretically the value should reflect at the time of execution. 
3. If we decide to have a inst member `VecConfig`, we can set it (a copy of global with some field override) at inst handler, and inst will carry it.  But agian, the member VecConfig is touched/updated each time at handler.

@kathlenemagnus  please let me now your opinion on this.

core/inst_handlers/v/RvvPermuteInsts.cpp
```diff
@@ -563,11 +563,12 @@ namespace pegasus
     Action::ItrType vmvrHelper(PegasusState* state, Action::ItrType action_it)
     {
         const PegasusInstPtr & inst = state->getCurrentInst();
-        VectorConfig config{*state->getVectorConfig()};
-        config.setLMUL(nReg * 8);
-        config.setVL(config.getVLMAX());
-        Elements<Element<elemWidth>, false> elems_vs2{state, &config, inst->getRs2()};
-        Elements<Element<elemWidth>, false> elems_vd{state, &config, inst->getRd()};
+        auto configPtr = inst->getVecConfig();
+        configPtr = std::make_shared<VectorConfig>(*state->getVectorConfig());
```
colby-nyce, line 567: I don't think this code is doing what you intended. Even though the `getVecConfig()` method returns the shared_ptr by reference, this line:

`auto configPtr = inst->getVecConfig();`

Still makes a copy.

```
#include <iostream>
#include <memory>

struct VecConfig {
    int x, y, z;
    VecConfig(int x, int y, int z) : x(x), y(y), z(z) {}
};

std::ostream& operator<<(std::ostream& os, const VecConfig& vec_config) {
    os << vec_config.x << ", " << vec_config.y << ", " << vec_config.z;
    return os;
}

struct Inst {
    std::shared_ptr<VecConfig> vec_config;
    std::shared_ptr<VecConfig> & getVecConfig() { return vec_config; }
};

int main() {
    Inst inst;
    inst.vec_config = std::make_shared<VecConfig>(1,2,3);
    auto vec_config = inst.getVecConfig();           // vec_config is a copy of inst.vec_config
    vec_config = std::make_shared<VecConfig>(4,5,6); // has no effect on inst.vec_config

    std::cout << "Orig: " << *inst.vec_config << std::endl;
    std::cout << "Copy: " << *vec_config << std::endl;
}
```

The above code prints out:

```
Orig: 1, 2, 3
Copy: 4, 5, 6
```

You need `auto & configPtr = inst->getVecConfig()` to persist changes to the inst's vec config.

But it's not clear why you are deleting the old config / creating a new one with every method call. What is the intention behind `std::make_shared` here?

include/PegasusTypes.hpp
```diff
@@ -112,6 +112,17 @@ namespace pegasus
...
     };
 
+    enum class RoundingMode
+    {
+        EVEN = 0,
+        MINMAG = 1,
+        MIN = 2,
+        MAX = 3,
+        MAXMAG = 4,
+        ODD = 6,
```
colby-nyce, line 122: Did you intend to skip over the enum value 5? Why not just use:

```
    enum class RoundingMode
    {
        EVEN,
        MINMAG,
        MIN,
        MAX,
        MAXMAG,
        ODD,
        DYN
    }

File: core/inst_handlers/v/RvvPermuteInsts.cpp (hunk 2 above)
Comment (lzhu-pub) [reply core/inst_handlers/v/RvvPermuteInsts.cpp:567]: @colby-nyce  Thanks for pointing this out!
Maybe I shall just add a new assignment operator for VecConfig. There is no need to deleting old VecConfig.

File: include/PegasusTypes.hpp (hunk 3 above)
Comment (lzhu-pub) [reply include/PegasusTypes.hpp:122]: Yes, I do intend to.

File: core/PegasusInst.hpp [hunk 4]
```diff
@@ -187,6 +193,9 @@ namespace pegasus
         // Cache immediate value, unsigned and signed
         const uint64_t immediate_value_;
 
+        // Vector Config that is different than global
+        std::shared_ptr<VectorConfig> vec_config_;
```
kathlenemagnus, line 197: Why not use `SpartaSharedPointer` class instead? See https://github.com/sparcians/map/blob/db25676dd79a1b0c528c90fa0069803bca88c309/sparta/sparta/utils/SpartaSharedPointer.hpp

PR #280, opened 2025-10-22
Moved Mavis and InstHandlers into PegasusCore

kathlenemagnus-mips:
Moved more core components into PegasusCore. It made sense to move the InstHandlers out of Execute and into PegasusCore because they need to be accessed by PegasusExtractor.


klingaard (approved):
Wow, nice cleanup.

PR #287, opened 2025-11-03
Merge event pipeline and checkpointer for CoSim

colby-nyce:
The "pegasus_cosim_regress" target runs all ISA tests against the cosim/flush harness. Current test results:
**PASS RATE: 86.67% (299/345)**

This is a couple weeks' worth of work so we don't have to wait for a higher pass rate to get this checked in. I will work on the failing tests and post smaller PRs for the fixes.

### DIFF
--- a/scripts/RunArchTests.py
+++ b/scripts/RunArchTests.py
@@ -213,6 +213,15 @@ def main():
     parser.add_argument("--expected-pass-rate", type=float, help="Expected pass rate (for CI purposes only)")
     args = parser.parse_args()
 
+    # Extract the RISC-V tarball if needed
+    if args.riscv_arch is not None:
+        if not os.path.isdir(args.riscv_arch):
+            tarball = args.riscv_arch + ".tar.gz"
+            if os.path.isfile(tarball):
+                print("Extracting RISC-V arch tests from", tarball)
+                subprocess.run(["tar", "-xzf", tarball], check=True)
+                args.riscv_arch = os.getcwd() + "/" + os.path.basename(args.riscv_arch)
+
     global be_noisy
 
     ###########################################################################


core/inst_handlers/zicsr/RvzicsrInsts.cpp
```diff
@@ -417,25 +417,49 @@ namespace pegasus
...
 
         auto & ext_manager = state->getCore()->getExtensionManager();
+        bool change_mavis_ctx = false;
+
         // If FS is set to 0 (off), all floating point extensions are disabled
         const uint32_t fs_val = READ_CSR_FIELD<XLEN>(state, MSTATUS, "fs");
         if (fs_val == 0)
         {
-            ext_manager.disableExtensions({"f", "d"});
+            std::vector<std::string> disabled_exts;
```
kathlenemagnus-mips: These changes will probably apply to the MISA update handler as well so you might want to consider moving some of this code to PegasusState. These tests do a lot of writes to the MISA CSR:

- rv64uc-p-rvc
- rv64uc-v-rvc

test/cosim/cosim_workload/flush_workload/FlushWorkload_test.cpp
```diff
@@ -0,0 +1,451 @@
...
+    return true;
+}
+
+template <typename XLEN>
+bool Compare(PegasusCoSim & sim_truth, PegasusCoSim & sim_test, CoreId core_id, HartId hart_id)
+{
+    auto state_truth = sim_truth.getPegasusCore(core_id)->getPegasusState(hart_id);
+    auto state_test = sim_test.getPegasusCore(core_id)->getPegasusState(hart_id);
+
+    // Compare PegasusState member variables
```
kathlenemagnus-mips: Would it be better to define an equality operator in the PegasusState class? I know it won't report the exact member variable that was different, but it may be easier to maintain since it'll be in the same class as the member variables.

test/cosim/cosim_workload/flush_workload/FlushWorkload_test.cpp
```diff
@@ -0,0 +1,451 @@
...
+        constexpr bool flush_younger_only = false;
+        sim.flush(event, flush_younger_only);
+    }
+
+    sim.commit(stepped_events.front());
+    return true;
+}
+
+template <typename XLEN>
+bool Compare(PegasusCoSim & sim_truth, PegasusCoSim & sim_test, CoreId core_id, HartId hart_id)
```
kathlenemagnus-mips: Wow! This function doesn't miss a thing! 

Once we hit a 100% pass rate with the RISC-V arch tests, we may want to consider removing some of the redundant checks. For example, we probably don't need to check the dasm string and the opcode. If the opcode is different, so will the dasm string.

test/cosim/cosim_workload/flush_workload/FlushWorkload_test.cpp
```diff
@@ -0,0 +1,451 @@
...
+int main(int argc, char** argv)
+{
+    const auto [workload, max_steps_before_flush, fast_forward_steps] = ParseArgs(argc, argv);
+    const auto arch = GetArchFromPath(workload);
+
+    // Disable sleeper thread so we can run two simulations at once.
+    sparta::SleeperThread::disableForever();
+
+    const uint64_t ilimit = 0;
+    const auto uuid = GenerateUUID();
```
kathlenemagnus-mips: Why not just use the wkld name?

cosim/CoSimEventPipeline.cpp
```diff
@@ -0,0 +1,810 @@
...
+    void CoSimEventPipeline::createPipeline(simdb::pipeline::PipelineManager* pipeline_mgr)
+    {
+        auto db_accessor = pipeline_mgr->getAsyncDatabaseAccessor();
+        auto pipeline = pipeline_mgr->createPipeline(NAME);
+
+        pipeline_mgr_ = pipeline_mgr;
+        async_eval_ = db_accessor;
+
+        // Task 1: Run the incoming events through boost::serialization
+        using EventList = std::deque<Event>;
```
kathlenemagnus-mips: Can this be moved to the header?

cosim/CoSimEventPipeline.cpp
```diff
@@ -0,0 +1,810 @@
...
+            last_event_uid_ = euid;
+            observer->event_uid_ = euid;
+            sim_stopped_ = reload_evt.isLastEvent();
+
+            state->setPc(reload_evt.getNextPc());
+            state->setPrivMode(reload_evt.getNextPrivilegeMode(), state->getVirtualMode());
+
+            auto sim_state = state->getSimState();
+            sim_state->reset();
+            sim_state->current_opcode = reload_evt.getOpcode();
```
kathlenemagnus-mips: I don't think it's necessary to reset the opcode since as soon as we call step again it's going to be cleared. Also keep in mind that `sim_state->reset()` only clears the opcode and the current inst ptr.

cosim/CoSimEventPipeline.cpp
```diff
@@ -0,0 +1,810 @@
...
+                state->getCore()->changeMavisContext();
+            }
+
+            // Update MMU mode when MSTATUS / SATP csrs have changed,
+            // or when the privilege mode(s) have changed.
+            change_mmu_mode |= evt.curr_priv_ != evt.next_priv_;
+            change_mmu_mode |= evt.curr_ldst_priv_ != evt.next_ldst_priv_;
+            if (!change_mmu_mode && evt.inst_csr_ != std::numeric_limits<uint32_t>::max())
+            {
+                change_mmu_mode |= evt.inst_csr_ == MSTATUS || evt.inst_csr_ == SATP;
```
kathlenemagnus-mips: I feel like this logic with MSTATUS and SATP should live somewhere else but I'm not sure where. Would be great if we could somehow ask the CSR if it can affect the MMU mode.

kathlenemagnus (approved):

test/cosim/cosim_workload/flush_workload/FlushWorkload_test.cpp
```diff
@@ -0,0 +1,451 @@
...
+#include <regex>
+
+/// In this test, we will be running the same workload through two PegasusCoSim
+/// instances. One will ONLY step forward and serve as the "truth" against which
+/// we will validate registers, PC, and other state. The other instance will step
+/// forward N times, then flush to the N-1 instruction. Then the comparison takes
+/// place, and we continue until the end of the workload.
+
+using namespace pegasus;
+using namespace pegasus::cosim;
```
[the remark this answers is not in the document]
colby-nyce: It wouldn't compile if there were collisions, but I'll change it so it is future-proof.

scripts/RunArchTests.py
```diff
@@ -127,6 +127,77 @@ def run_tests_in_parallel(tests, passing_tests, failing_tests, timeout_tests, ou
...
+                test_expr, _, _ = match.groups()
+                similar_failures = failure_dict.get(test_expr, set())
+                similar_failures.add(os.path.basename(log_file))
+                failure_dict[test_expr] = similar_failures
+
+
+def print_sparta_failures():
+    failure_dict = {}
+    for entry in os.listdir(os.getcwd()):
+        if entry.endswith('.log'):
```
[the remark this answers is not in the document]
colby-nyce: Sparta writes the error logs when it catches an exception. The harness already does try/catch and if an exception is caught, the check for `EXPECT_EQUAL(exception_str, "")` will fail, then the harness return code is non-zero. So RunArchTests.py will see that the test failed and it will leave `<workload>.log` in the working directory. I don't think we have a need for reporting the backtrace too which is all that these changes would give us.

test/cosim/cosim_workload/flush_workload/FlushWorkload_test.cpp
colby-nyce: It's helpful to be able to run the whole suite and see individual failure conditions. As a compromise, I think we can add something like this to PegasusState, PegasusCore, PegasusInst, SimState, and VectorConfig:

template <bool IS_UNIT_TEST = false>
bool compare(const PegasusState* state) const;

Then use "if constexpr" to decide to use EXPECT_EQUAL.

When the pass rate is 100% I'll go back and make this "bool operator==()" instead.

cosim/PegasusCoSim.cpp
```diff
@@ -101,9 +117,23 @@ namespace pegasus::cosim
...
+                ++pipeline_idx;
+
+                auto cosim_obs = std::make_unique<CoSimObserver>(cosim_logger_, evt_pipeline,
+                                                                 checkpointer, core_idx, hart_idx);
+
+                evt_pipeline->setObserver(cosim_obs.get());
+
+                // Initialize PegasusState and take initial snapshot
+                state->boot();
+                checkpointer->setSnapshotThreshold(10);
```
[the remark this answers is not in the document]
colby-nyce: Fixed

core/PegasusState.hpp
```diff
@@ -130,6 +135,7 @@ namespace pegasus
             {
                 current_opcode = 0;
                 current_inst.reset();
+                ++current_uid;
```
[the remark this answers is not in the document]
colby-nyce: We would actually have three IDs at that point: current_uid, arch_id, and inst_count. If current_uid is the one that is not reverted during flush, how is arch_id different than inst_count? Or if inst_count is also supposed to be always increasing, then how is current_uid different than inst_count? I feel like we are going to have duplicate IDs that track the same thing

PR #289, opened 2025-11-03
STF tracing support for RV32 

### LINKED ISSUE #281: Add RV32 STF Trace Generation Support

skhanx86:
Fixes #281


core/observers/STFLogger.cpp
```diff
@@ -41,100 +41,68 @@ namespace pegasus
         stf_writer_.setISA(stf::ISA::RISCV);
         stf_writer_.setHeaderPC(inital_pc);
         stf_writer_.finalizeHeader();
-        recordRegState_(state);
+        if (state->getXlen() == 32)
```
colby-nyce: Please use curly braces for one-liners to stay consistent with the rest of the Pegasus code.

core/observers/STFLogger.cpp
```diff
@@ -267,12 +237,82 @@ namespace pegasus
...
+        for (const auto & mem_read : mem_reads_)
+        {
+            stf_writer_ << stf::InstMemAccessRecord(mem_read.addr, mem_read.size, 0,
+                                                    stf::INST_MEM_ACCESS::READ);
+            stf_writer_ << stf::InstMemContentRecord(mem_read.value);
+        }
+
+        auto get_stf_reg_type = [](const RegType reg_type)
+        {
+            if (reg_type == RegType::INTEGER)
```
colby-nyce: Use a switch statement instead of if/elseif/elseif/...

core/observers/STFLogger.cpp
```diff
@@ -267,12 +237,82 @@ namespace pegasus
...
+            stfWrite_InstRegRecord_<uint32_t>(state, get_stf_reg_type);
+        }
+        else
+        {
+            stfWrite_InstRegRecord_<uint64_t>(state, get_stf_reg_type);
+        }
+
+        bool invalid_opcode = false;
+
+        // TODO: If the stf lib provides support for stf::EventRecord to contain 32 bit value, the
```
kathlenemagnus-mips: Does stf::EventRecord only support 64-bit opcodes? If so, can this comment be removed?

core/observers/STFLogger.cpp
```diff
@@ -41,100 +41,68 @@ namespace pegasus
...
+                                                      // uint64_t of dataPackets for vector.
             }
         }
 
         for (const auto & [csr_num, csr_read] : csr_reads_)
         {
             stf_writer_ << stf::InstRegRecord(csr_num, stf::Registers::STF_REG_TYPE::CSR,
                                               stf::Registers::STF_REG_OPERAND_TYPE::REG_SOURCE,
-                                              csr_read.getRegValue<uint64_t>());
+                                              csr_read.template getRegValue<XLEN>());
```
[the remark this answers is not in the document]
skhanx86: The compiler throws an error saying: `expected 'template' keyword before dependent template name`, hence I added it.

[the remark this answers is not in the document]
kathlenemagnus-mips: I found this read very interesting: https://www.en.cppreference.com/w/cpp/language/dependent_name.html

Look for section "**The template disambiguator for dependent names**" that explains why this keyword is needed. The type of `getRegValue` is dependent on the value of XLEN, so it can't be known until the value of the XLEN is known at instantiation. So I think the keyword is needed to tell the compiler that this function call's type cannot be evaluated until instantiation.

PR #297, opened 2025-11-14
Added PegasusSimParameters Extension

kathlenemagnus-mips:
Created an extension to store the command line and simulation-level parameters. 

You can now do this:
```
./pegasus --write-final-config rerun.yaml --reg "core0.hart0.sp 0x0000003ffffff000" --reg "core0.hart0.gp 0x77000" --reg "core0.hart0.tp 0x7d000" -p top.extension.sim.enable_syscall_emulation true ../test/sim/workloads/dhry.elf
```
```
./pegasus -c rerun.yaml
```

The `rerun.yaml` will contain all of the Pegasus parameters and the new extension parameters:
```
top:
  extension.sim:
    enable_syscall_emulation: true
    workloads: "[[/home/kathlenemagnus/forks/pegasus/test/sim/workloads/dhry.elf]]"
    inst_limit: 0
    reg_overrides: "[[core0.hart0.sp, 0x0000003ffffff000],[core0.hart0.gp, 0x77000],[core0.hart0.tp, 0x7d000]]"
```

### DIFF
--- a/test/cosim/cosim_workload/flush_workload/FlushWorkload_test.cpp
+++ b/test/cosim/cosim_workload/flush_workload/FlushWorkload_test.cpp
@@ -335,9 +335,9 @@ int main(int argc, char** argv)
     config_truth.enableLogging("top", "inst", workload_fname + ".log");
     config_truth.processParameter("top.core0.params.isa", "rv64gcbv_zicsr_zifencei_zicond_zfh",
                                   false);
-    PegasusSimParameters::WorkloadsAndArgs workloads_and_args{{workload}};
+    pegasus::PegasusSimParameters::WorkloadsAndArgs workloads_and_args{{workload}};
     const std::string wkld_param =
-        PegasusSimParameters::convertVectorToStringParam(workloads_and_args);
+        pegasus::PegasusSimParameters::convertVectorToStringParam(workloads_and_args);
     config_truth.processParameter("top.extension.sim.workloads", wkld_param);
     config_truth.processParameter("top.extension.sim.inst_limit", std::to_string(ilimit));
     config_truth.copyTreeNodeExtensionsFromArchAndConfigPTrees();


sim/pegasus.cpp
```diff
@@ -81,42 +82,71 @@ int main(int argc, char** argv)
...
+        }
+
+        // Process command line parameters
+        auto & sim_cfg = cls.getSimulationConfiguration();
+
+        // Workload
+        if (workloads.empty() == false)
         {
-            if (workload.empty() && (0 == vm.count("opcode")))
+            std::string wkld_and_args_param_value = "[";
```
kathlenemagnus: I know this is very wonky. I'll clean it up on Monday.

sim/PegasusSim.cpp
```diff
@@ -138,48 +133,78 @@ namespace pegasus
...
+        for (const auto & reg_override : reg_overrides)
+        {
+            sparta_assert(reg_override.size() == 2, "Register override param is malformed!");
+            std::vector<std::string> tmp;
+            const std::string delim = ".";
+            sparta::utils::tokenize(reg_override[0], tmp, delim);
+            sparta_assert(tmp.size() == 3, "Register override param is malformed!");
+            const std::string & core = tmp[0];
+            const std::string & hart = tmp[1];
+            const std::string & reg_name = tmp[2];
```
klingaard: I'm a little confused -- this is looking for a register override in the format:
```
[core*.hart*.regname, value]
```
But I don't think this the format in which the register overrides are stored:
```
[regname, value]
```

klingaard (approved):
Preemptive approval

sim/PegasusSim.cpp
kathlenemagnus-mips: "regname" is just a string that needs to be split and parsed. I can rename the variable to just "reg".

klingaard: What I guess I'm saying is that the code appears to be expecting `core` and `hart` in the name, but the code above does not store that.

Issue #282, opened 2025-10-27
Add Vector STF Trace Generation Support


kathlenemagnus:
- [ ] Consider the LMUL/EMUL when capturing source and destination registers
- [x] Capture VL and VTYPE CSRs for all vector instructions
- [ ] Capture the mask register V0 for masked vector instructions

PR #299, opened 2025-11-21
Fix Vector Register Tracing for LMUL/EMUL and Masked Instructions (#282)

skhanx86:
This PR implements two fixes to address incorrect vector register tracing as described in Issue #282:

1. Correct LMUL/EMUL handling
- Source and destination vector registers are now captured based on the actual LMUL/EMUL setting, ensuring the correct number of physical vector registers are traced for both fractional and widened LMUL values.

2. Capture mask register V0 for masked vector instructions
- Masked vector instructions (VM = 0) now correctly trace V0 as a source register. Instructions without a VM field are safely ignored.

Fixes: #282

### DIFF
--- a/core/observers/Observer.hpp
+++ b/core/observers/Observer.hpp
@@ -141,7 +141,7 @@ namespace pegasus
         {
             using ObservedReg::ObservedReg;
 
-            //store LMUL-wide register values // only for sources
+            // store LMUL-wide register values // only for sources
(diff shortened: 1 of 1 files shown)


skhanx86:
@kathlenemagnus as mentioned in comment above, the destination registers need not pre-store the value as it can be fetched directly at the time of writing to the STF buffer, no fix for the destination register is required for STF tracing.

PR #300, opened 2025-12-03
Implemented Pause Instruction and Added Multithreaded Test to Regression

kathlenemagnus-mips:
Added support for the pause instruction, which stops the execution of a thread until a pause counter expires. The counter's duration is in cycles and can be set by a parameter. Also added a true multithreaded test case to the regression, which requires a few miscellaneous fixes to get working properly.

### DIFF
--- a/core/PegasusInst.hpp
+++ b/core/PegasusInst.hpp
@@ -125,6 +125,36 @@ namespace pegasus
             return opcode_info_->getSpecialField(mavis::OpcodeInfo::SpecialField::VM);
         }
 
+        bool isVectorInstMasked() const
+        {
+            try
+            {
+                // In RISC-V V-extension:
+                //   VM = 0 → masked operation (uses v0.t)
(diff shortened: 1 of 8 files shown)


klingaard (approved):

core/PegasusCore.cpp
```diff
@@ -296,10 +303,16 @@ namespace pegasus
...
+    // This method will execute a single thread for up to X instructions
+    // where X is the quantum for that thread. The cycle count of all threads
+    // is updated to match the current thread's cycle count. Then this event
+    // is rescheduled at that cycle count.
     void PegasusCore::advanceSim_()
     {
         PegasusState* state = threads_[current_hart_id_];
-        if (state->getSimState()->sim_stopped == false)
+        bool pause_thread = false;
+        if ((state->getSimState()->sim_stopped == false)
```
klingaard: Might be worth making a local variable to cache the simstate since `state->getSimState()` is called a lot.

PR #301, opened 2025-12-10
Implemented CoSim Register Access Methods

kathlenemagnus:
Implemented the CoSim read/peek/write/poke register methods and added new methods. Also aligned the RegType enum values and the Sparta register group numbers and made a couple of other CoSim Observer fixes.


cosim/PegasusCoSim.cpp
```diff
@@ -381,6 +484,36 @@ namespace pegasus::cosim
...
+        const size_t size = buffer.size();
+        const size_t OFFSET = 0;
+        reg->write(buffer.data(), size, OFFSET);
+    }
+
+    void PegasusCoSim::pokeRegister_(sparta::Register* reg, std::vector<uint8_t> & buffer) const
+    {
+        const size_t size = buffer.size();
+        const size_t OFFSET = 0;
+        reg->write(buffer.data(), size, OFFSET);
```
lzhu-pub: is it supposed to be `poke`?

kathlenemagnus: Yep! Thanks for catching that.

klingaard (approved):

PR #310, opened 2026-01-08
Record Memory Reads and Writes in CoSim Event

kathlenemagnus-mips:
The CoSim Observer wasn't populating the Event memory read and writes. Hooking everything up required new read/write memory method in PegasusState to be able to observe the source of the memory access and the virtual address.


system/PegasusSystem.cpp
```diff
@@ -327,7 +327,6 @@ namespace pegasus
 
     void PegasusSystem::registerMemoryCallbacks(Observer* observer)
     {
-        observer->registerReadWriteMemCallbacks(getSystemMemory());
```
kathlenemagnus-mips: @colby-nyce can you double-check what I did here makes sense? I was seeing all reads and writes being observed twice and this fixed the problem. I think the system memory is already included in `tree_nodes_`.

kathlenemagnus-mips:
This is a WIP because I still need to update all of the instruction handlers to use the new read/write memory methods so they are observed correctly.

I also need to fix the cosim regress target. Only 53.04% of the tests are passing.

cosim/Event.hpp
```diff
@@ -108,6 +108,18 @@ namespace pegasus::cosim
             size_t size;
             std::vector<uint8_t> value;
 
+            MemReadAccess() = default;
+
+            MemReadAccess(MemAccessSource source, Addr paddr, Addr vaddr, size_t size,
+                          const std::vector<uint8_t> value) :
```
colby-nyce: Pass by const ref

cosim/Event.hpp
```diff
@@ -125,6 +137,16 @@ namespace pegasus::cosim
         {
             std::vector<uint8_t> prev_value;
 
+            MemWriteAccess() = default;
+
+            MemWriteAccess(MemAccessSource source, Addr paddr, Addr vaddr, size_t size,
+                           const std::vector<uint8_t> value,
```
colby-nyce: Pass by const ref

system/PegasusSystem.cpp
colby-nyce: Do you remember which ISA test you were bashing with? If I add this code at the end of createMemoryMappings_, I get the exception for rv64mi-p-csr:

```
        auto iter = std::find_if(
            tree_nodes_.begin(), tree_nodes_.end(),
            [this](const std::unique_ptr<sparta::TreeNode> & tnode) {
                return tnode.get() == memory_map_.get();
            });

        if (iter == tree_nodes_.end())
        {
            throw sparta::SpartaException()
                << "ERROR: PegasusSystem memory_map_ not found in tree_nodes_!";
        }
```

I'm wondering if sometimes it's in tree_nodes_ and sometimes not. Can this be added to registerMemoryCallbacks?

```
        auto iter = std::find_if(
            tree_nodes_.begin(), tree_nodes_.end(),
            [this](const std::unique_ptr<sparta::TreeNode> & tnode) {
                return tnode.get() == memory_map_.get();
            });

        if (iter == tree_nodes_.end())
        {
            observer->registerReadWriteMemCallbacks(getSystemMemory());
        }
```

core/observers/Observer.hpp
```diff
@@ -177,16 +179,72 @@ namespace pegasus
...
+            ObservedMemAccess(const ObservedMemAccess & other) :
+                paddr(other.paddr),
+                vaddr(other.vaddr),
+                size(other.size),
+                mem_value(other.mem_value),
+                source(other.source)
+            {
+            }
+
+            ObservedMemAccess & operator=(ObservedMemAccess other)
```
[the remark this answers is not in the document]
kathlenemagnus: I implemented it myself because I was getting this compiler error:
```
error: explicitly defaulted copy assignment operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
  202 |             ObservedMemAccess & operator=(ObservedMemAccess& other) = default;
      |                                 ^
/home/kmagnus/forks/pegasus/core/observers/Observer.hpp:206:24: note: copy assignment operator of 'ObservedMemAccess' is implicitly deleted because field 'paddr' is of const-qualified type 'const Addr' (aka 'const unsigned long')
  206 |             const Addr paddr;
      |                        
```

Is there a better workaround for this?

Issue #313, opened 2026-01-12
Set correct ISA string during STF generation to avoid post-processing with stf_isa_edit


skhanx86:
Currently, Pegasus generates STF files with a default (which can be incomplete for a few cases) ISA string. As a result, users must run an additional post-processing step using the ISA-isa-edit tool to fix the ISA metadata before the STF can be consumed downstream.

This extra step is unnecessary and can be avoided if Pegasus sets the correct ISA string at STF generation time.

PR #314, opened 2026-01-12
Fix ISA string in STF generation to eliminate isa-edit post-processing

skhanx86:
This PR updates Pegasus to set the correct ISA string during STF generation. Previously, STF files required post-processing with the stf_isa_edit tool to avoid stf_check errors. With this change, generated STFs are immediately valid and should pass stf_check without additional steps. (Fixes: #313)

### DIFF
--- a/cosim/PegasusCoSim.cpp
+++ b/cosim/PegasusCoSim.cpp
@@ -55,25 +55,26 @@ namespace pegasus::cosim
     }
 
     PegasusCoSim::PegasusCoSim(uint64_t ilimit, const std::string & workload,
+                               const std::map<std::string, std::string> pegasus_params,
                                const std::string & db_file, const size_t snapshot_threshold)
     {
-        // TODO: Assume 1 core, 1 hart for now
-        const uint32_t num_cores = 1;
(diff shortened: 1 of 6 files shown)


core/observers/STFLogger.cpp
```diff
@@ -41,8 +42,11 @@ namespace pegasus
         // TODO: Add support for PTE
         // stf_writer_.setTraceFeature(stf::TRACE_FEATURES::STF_CONTAIN_PTE);
 
+        auto isa = state->getCore()->getISAString();
```
kathlenemagnus: This can be `const`, too.

core/observers/STFLogger.cpp
```diff
@@ -41,8 +42,11 @@ namespace pegasus
         // TODO: Add support for PTE
         // stf_writer_.setTraceFeature(stf::TRACE_FEATURES::STF_CONTAIN_PTE);
 
+        const auto isa = state->getCore()->getISAString();
```
bdutro: Make this `const auto&` to avoid another copy.

PR #317, opened 2026-01-16
Add Support for Loading Multiple ELFs

kathlenemagnus:
Came across a use case for Pegasus where the kernel/boot code was stored in an ELF and the workload was stored in a separate ELF. Modified PegasusSystem to support loading multiple ELFs. The first ELF is seen as the "primary" ELF and will set the starting PC.

### DIFF
--- a/core/inst_handlers/v/RvvFloatInsts.cpp
+++ b/core/inst_handlers/v/RvvFloatInsts.cpp
@@ -1245,15 +1245,28 @@ namespace pegasus
                 auto index = iter.getIndex();
                 if constexpr (opMode.src1 == OperandMode::Mode::V)
                 {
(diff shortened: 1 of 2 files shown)


system/PegasusSystem.hpp
```diff
@@ -49,7 +49,7 @@ namespace pegasus
         void registerMemoryCallbacks(Observer* observer);
 
         // Get starting PC from ELF
-        Addr getStartingPc() const { return starting_pc_; }
+        Addr getStartingPc() const { return starting_pc_.isValid() ? starting_pc_.getValue() : 0; }
```
klingaard: Wonder if we should add a parameter to force-set the starting PC instead of using 0.

klingaard (approved):

PR #328, opened 2026-02-08
Added zknd implementation

Ma-gi-cian:
I have used the `riscv-crypto-spec-scalar` pdf and the Whisper's `crypto.cpp` and  `crypto-util.hpp` for this. The project successfully builds will also create the pr for the mavis part.

open to any feedback, corrections, style suggestions, or improvements — happy to iterate on them. Thanks.

### DIFF
--- a/core/inst_handlers/zknd/RvzkndInsts.cpp
+++ b/core/inst_handlers/zknd/RvzkndInsts.cpp
@@ -70,7 +70,6 @@ namespace pegasus
         const XLEN rs1_val = READ_INT_REG<XLEN>(state, inst->getRs1());
         const uint32_t imm = inst->getImmediate();
         const XLEN rd_val = OPERATOR{}(rs1_val, 0, imm);
-        printf("rs1_val : %lu, imm : %d and the result is : %lu", rs1_val, imm, rd_val);
         WRITE_INT_REG<XLEN>(state, inst->getRd(), rd_val);
         return ++action_it;
     }


core/inst_handlers/zknd/RvzkndInsts.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+        uint64_t output = 0;
+        for (int i = 0; i < 8; i++)
+        {
+            output |= (uint64_t)result[i] << (i * 8);
+        }
+        return output;
+    }
+
+
```
kathlenemagnus: Delete extra empty line.

core/inst_handlers/zknd/RvzkndInsts.hpp
```diff
@@ -0,0 +1,61 @@
...
+        using base_type = RvzkndInsts;
+
+        template <typename XLEN>
+        static void getInstHandlers(InstHandlers::InstHandlersMap & inst_handlers);
+
+      private:
+        template <typename XLEN, typename OPERATOR>
+        Action::ItrType aesHandler_(PegasusState* state, Action::ItrType action_it);
+
+        template <typename XLEN> struct Aes32DsiOp
```
kathlenemagnus: Can you move the functors into a separate header called `RvzkndFunctors.hpp`? I think some of these functors can be reused by the vector crypto instructions whenever we get around to implementing those. Putting them in a separate header (and outside of the `RvzkndInsts`) class will make them easier to reuse across extensions.

PR #333, opened 2026-02-12
CMake updates to make integrating Pegasus into other projects easier

bdutro-mips:
Also includes changes to support latest STF and Mavis


CMakeLists.txt
```diff
@@ -153,63 +157,71 @@ FetchContent_MakeAvailable(spike)
...
+endif()
 
 ################################################################################
 
 # Install
 install(FILES ${PROJECT_BINARY_DIR}/softfloat/libsoftfloat.a DESTINATION lib)
-install(FILES ${PROJECT_BINARY_DIR}/stf_lib/lib/libstf.a DESTINATION lib)
-install(FILES ${PROJECT_BINARY_DIR}/mavis/libmavis.a DESTINATION lib)
+
+if (NOT PEGASUS_USE_PARENT_STF_LIB)
```
kathlenemagnus: Can you add some documentation in the README about building Pegasus as a part of other projects? Why and how you would use these CMake variables.

PR #335, opened 2026-02-13
fix lr/sc and reservation

lzhu-pub:
Fixed "lr sc" instruction handling.
Now the code is doing exactly what the comment indicated.
Elf test needs to be gdb'd to ensure the correct sequence is executed. Passing the test itself cannot guarantee correctness.

### DIFF
--- a/core/inst_handlers/a/RvaInstsBase.hpp
+++ b/core/inst_handlers/a/RvaInstsBase.hpp
@@ -111,16 +111,20 @@ namespace pegasus
         auto xlation_state = inst->getTranslationState();
 
         XLEN fail_code = 1; // assume bad
+
+        // From RISC-V spec:
+        //
+        // An SC may succeed only if no store from another hart to the
+        // reservation set can be observed to have occurred between the LR and the SC,
+        // and if there is no other SC between the LR and itself in program order.
         if (const auto & resv = state->getCore()->getReservation(state->getHartId());
-            resv.isValid())
+            resv.isValid() && resv == xlation_state->getResult().getPAddr()
+            && !state->storeOnReservationSetOccurred())
         {
-            if (resv == xlation_state->getResult().getPAddr())
-            {
-                const uint64_t rs2_val = READ_INT_REG<XLEN>(state, inst->getRs2());
-                state->writeMemory<SIZE>(xlation_state->getResult(), rs2_val,
-                                         MemAccessSource::INSTRUCTION);
-                fail_code = 0;
-            }
+            const uint64_t rs2_val = READ_INT_REG<XLEN>(state, inst->getRs2());
+            state->writeMemory<SIZE>(xlation_state->getResult(), rs2_val,
+                                     MemAccessSource::INSTRUCTION);
+            fail_code = 0;
         }
         xlation_state->popResult();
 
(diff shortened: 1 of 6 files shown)


core/inst_handlers/a/RvaInstsBase.hpp
```diff
@@ -114,7 +114,12 @@ namespace pegasus
         if (const auto & resv = state->getCore()->getReservation(state->getHartId());
             resv.isValid())
         {
-            if (resv == xlation_state->getResult().getPAddr())
+            // From RISC-V spec:
+            //
+            // An SC may succeed only if no store from another hart to the
+            // reservation set can be observed to have occurred between the LR and the SC,
+            // and if there is no other SC between the LR and itself in program order.
+            if (resv == xlation_state->getResult().getPAddr() && !state->storeOnReservationSetOccurred())
```
knute-mips: Is the address masked at all or is it bit-for-bit?

knute-mips (commented):
Hmmm... not sure about this.  @lzhu-mips I thought we chatted about using a custom memory interface to track the reservations and using that custom interface to capture the reads/writes and check for reservations?

Notifications work, but if we are to use DMI, this won't.  😦

lzhu-pub:
> > @knute-mips what is your suggestion on the memory wrapper approach?
> 
> Currently all accesses to memory from loads/stores/fetches go through this call in PegasusState:
> 
> ```c++
>         auto* memory = pegasus_core_->getSystem()->getSystemMemory();
> ```
> 
> The memory object being returned is a `sparta::memory:: BlockingMemoryIF` object, the derived type being a `sparta::memory:: SimpleMemoryMapNode`. What `memory` really points to means nothing to the core.
> 
> If `getSystemMemory()` returns a `sparta::memory:: BlockingMemoryIF` instead, then you can have the system have a pointer to what the latest memory interface is. For example, change this:
> 
> ```c++
>     // Get pointer to system memory
>     sparta::memory::SimpleMemoryMapNode* getSystemMemory() { return memory_map_.get(); }
> ```
> 
> to this:
> 
> ```c++
>     // Get pointer to system memory
>     sparta::memory::BlockingMemoryIF* getSystemMemory() { memory_view_; }
> ```
> 
> and have `memory_view_` initially set to `memory_map_.get();`.
> 
> When a reservation is set, adjust `memory_view_` to a new class called something like `ReservationMemory` that implements the read/write/getDMI/etc calls and holds the reservation value. "Insert" this memory object before the system memory.
> 
> Brain-dead implementation (this is NOT how you should do it, but to give an idea):
> 
> ```c++
>     void PegasusState::setResrvation(address) {
>           reservation_memory_->setAddress(address);
>           reservation_memory_->setSystemMemory(pegasus_core_->getSystem()->getSystemMemory());
>           pegasus_core_->getSystem()->setSystemMemory(reservation_memory_);
>      }
> ```
> 
> `getDMI` would be implemented as "fail." We'd have to also make calls to invalidate all DMI pointers, but that's now easy to add.

DMI pointers might be a little tricky here. So if we choose to "invalidate" any previously returned DMI, (which we should do since the old DMI might bypass our reservation registry), we are basically telling the code using DMI: "whatever I told you is invalid now, you figure out other ways to adjust to use the new `memory_view`"!
@kathlenemagnus  we haven't started using DMI in Pegasus yet, right?

knute-mips:
The rule (at least in SystemC) is when you broadcast a DMI invalidation, the holder must go through the front door again to get memory.  I.e. the calling code will go back to the system to get a pointer to memory again and ask it, "Can I get a DMI?  You invalidate the last one."  When the reservation is set the answer is "no."

PR #338, opened 2026-02-20
Added support for mtvec.MODE

khoipkv:
Add support for mtvec.MODE
- Turn mtvec.MODE writable in `scripts/RV32_CSR.py` and `scripts/RV64_CSR.py`.
- Added `mtvecUpdateHandler_`, which will check if the written mode is supported. If not, the write will be dropped to 0.
- `supported_trap_modes` is a vector of supported modes.

### DIFF
--- a/core/Exception.cpp
+++ b/core/Exception.cpp
@@ -123,15 +123,16 @@ namespace pegasus
         // VS-mode
         if (virt_mode && (priv_mode == PrivMode::SUPERVISOR))
         {
-            const TrapVectorMode vstvec_mode = READ_CSR_FIELD<XLEN>(state, VSTVEC, "mode");
-            const XLEN vstvec_base = READ_CSR_FIELD<XLEN>(state, VSTVEC, "base");
+            const TrapVectorMode vstvec_mode =
+                (TrapVectorMode)READ_CSR_FIELD<XLEN>(state, VSTVEC, "mode");
+            const XLEN vstvec_base = READ_CSR_FIELD<XLEN>(state, VSTVEC, "base") << 2;
             if (vstvec_mode == TrapVectorMode::DIRECT)
             {
                 trap_handler_address = vstvec_base;
             }
             if (vstvec_mode == TrapVectorMode::VECTORED)
             {
-                trap_handler_address = vstvec_base + 4 * cause_val;
+                trap_handler_address = vstvec_base + 4 * cause_val * (XLEN)is_interrupt;
             }
 
             WRITE_CSR_REG<XLEN>(state, VSEPC, epc_val);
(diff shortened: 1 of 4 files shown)


kathlenemagnus:
@khoipkv this is a great start! Your code is very clean.

What's missing is the actual change to trap handling when the MODE bit is changed to vectored mode. Here is where the PC is changed in our trap handling function:
https://github.com/sparcians/pegasus/blob/4f3a64248aaa4b1c3e7f52f352a74cd8dbe3cb28/core/Exception.cpp#L216

If the MODE is vectored and the trap is an interrupt, then set the PC to `trap_handler_address + (4*excp_code)`.

Also I would recommend creating an enum for the trap mode in [PegasusTypes.hpp](https://github.com/sparcians/pegasus/blob/main/include/PegasusTypes.hpp). Take a look at the enum `PrivMode` and this code for inspiration:
https://github.com/sparcians/pegasus/blob/4f3a64248aaa4b1c3e7f52f352a74cd8dbe3cb28/core/PegasusState.cpp#L38C1-L56C6

core/PegasusCore.hpp
```diff
@@ -84,6 +86,16 @@ namespace pegasus
             return supported_priv_modes_.contains(mode);
         }
 
+        bool isTrapModeSupported(const int mode) const
+        {
+            for (int m : supported_trap_modes_)
```
colby-nyce: You can collapse this loop to just:

```
const auto & modes = supported_trap_modes_;
return std::find(modes.begin(), modes.end(), mode) != modes.end());
```

Or:

```
const auto & modes = supported_trap_modes_;
return std::any_of(modes.begin(), modes.end(), [=](int m) { return m == mode; });
```

If you leave the code as-is, please use curly braces:

```
if (m == mode)
{
    return true;
}
```

core/Exception.cpp
```diff
@@ -123,7 +123,17 @@ namespace pegasus
         // VS-mode
         if (virt_mode && (priv_mode == PrivMode::SUPERVISOR))
         {
-            trap_handler_address = (READ_CSR_REG<XLEN>(state, VSTVEC) & ~(XLEN)1);
+            const TrapVectorMode vstvec_mode =
+                (TrapVectorMode)READ_CSR_FIELD<XLEN>(state, VSTVEC, "mode");
+            const XLEN vstvec_base = READ_CSR_FIELD<XLEN>(state, VSTVEC, "base") << 2;
+            if (vstvec_mode == TrapVectorMode::DIRECT)
+            {
+                trap_handler_address = vstvec_base;
```
kathlenemagnus: Small nitpick but you just need 1 if-statement:
```
trap_handler_address = vstvec_base;
if (vstvec_mode == TrapVectorMode::VECTORED)
{
    trap_handler_address += 4 * cause_val * (XLEN)is_interrupt;
}
```

PR #342, opened 2026-02-26
Fixed zknd

Ma-gi-cian:
**ZKND FIX**

1. `aes32dsi/aes32dsmi`: replaced left shift with `std::rotl` to faithfully match the SAIL spec's `rol32`. The original used bit insertion (following whisper's implementation) which produces identical results for these operands, but rotl is the more accurate translation.
2. `aes64ks1i`: was incorrectly using `aes_sbox_inv` instead of `aes_sbox_fwd` for the SubWord step in the key schedule.
3. `aes64ks1i` now uses a dedicated `aesKsiOpHandler_` with `rnum` > 10 illegal instruction check per spec.

Testing is currently blocked --  the mavis JSON files for zknd are incorrect. They were based on existing JSON files as a template, and when opcode collisions occurred, masks were added as a workaround which appeared to work when generating golden.out but testing it has not been possible.

I also attempted using the overlay mechanism from the mavis JSON instruction definition docs to resolve the collisions:

```json
"overlay" : {
    "base" : "<colliding_mnemonic>",
    "match" : ["0x<mask>", "0x<opcode>"]
}
```

However, the same decoding issue persists and I was not able to resolve it. If someone could guide me on the correct approach for defining the zknd instruction encodings in mavis JSON, that would be greatly appreciated.

Note: I apologize for the incomplete initial implementation. The mavis JSON files were written by following existing file patterns, but when opcode collisions occurred I added masks as a workaround without fully understanding the encoding. This was the wrong approach and the JSON files need to be corrected properly before end-to-end testing can be completed. Will be sure to make prs as draft until the entire testing has not been completed.


kathlenemagnus (approved):

Ma-gi-cian:
@kathlenemagnus could you trigger the merge I do not have the permission for it. Thanks.

PR #366, opened 2026-03-25
Add smrnmi extension basic support

knute-mips:
CSRs in place, `mnret` stubbed out, not implemented.

closes #364

### DIFF
--- a/cosim/PegasusCoSim.cpp
+++ b/cosim/PegasusCoSim.cpp
@@ -596,19 +596,19 @@ namespace pegasus::cosim
 
     void PegasusCoSim::finish()
     {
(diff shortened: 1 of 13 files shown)


cosim/PegasusCoSim.cpp
knute-mips: @colby-nyce -- to get regression to pass on MacOS, I had to rearrange the code below.  I was getting an assert on line 608 (invalid event UID).  I think we have to run the SimLoopTeardown _after_ we query the events?

kathlenemagnus (approved):
LGTM

colbynyce-mips (approved):

colby-nyce (approved):

PR #373, opened 2026-04-10
Implemented Zicntr and Zihpm Extension

kathlenemagnus:
Added support for the Zicntr (cycle, time and instret) and Zihpm extensions.


core/PegasusState.cpp
```diff
@@ -524,10 +524,17 @@ namespace pegasus
 
         // Increment instruction count
         ++sim_state_.inst_count;
+        ++cached_csrs_.at(INSTRET);
+        ++cached_csrs_.at(MINSTRET);
```
klingaard: So are these cached csrs only available if ZIcntr is enabled?  In other words, if that extension is disabled, this code will crash, right?

core/PegasusState.hpp
```diff
@@ -410,6 +415,11 @@ namespace pegasus
         //! Simulation state
         SimState sim_state_;
 
+        //! Cached CSRs, indexed by CSR number
+        std::map<uint32_t, uint64_t> cached_csrs_{
+            {CYCLE, 0}, {MCYCLE, 0}, {TIME, 0}, {INSTRET, 0}, {MINSTRET, 0},
+        };
```
klingaard: Since this is hard-coded, does it make sense to have a map?  Why not just break out simple variables?

core/PegasusState.cpp
kathlenemagnus: I removed the cached CSR values and decided to just increment all of the counters every cycle. With the DMI read/write methods, it shouldn't slow down performance too much.

core/PegasusState.cpp
```diff
@@ -511,7 +525,91 @@ namespace pegasus
         }
     }
 
-    template <bool CHECK_ILIMIT>
+    template <typename XLEN> void PegasusState::incrCycleCsrs_()
```
kathlenemagnus, line 609: These methods need to check that the Zicntr extension is enabled.

klingaard (approved):

kathlenemagnus:
@klingaard FYI I completely reworked this PR after you reviewed it so please take another look!

kathlenemagnus:
This PR really needs @elliottgurrola-mips's PR #368 to work properly. Right now, the counter CSRs are always enabled even if the Zicntr/Zihpm extensions are not.

core/PegasusState.cpp
```diff
@@ -511,7 +525,91 @@ namespace pegasus
...
+    template <typename XLEN> void PegasusState::incrInstretCsrs_()
+    {
+        XLEN instret = getCsrRegister(INSTRET)->dmiRead<XLEN>() + 1;
+        XLEN minstret = getCsrRegister(MINSTRET)->dmiRead<XLEN>() + 1;
+        if constexpr (std::is_same_v<XLEN, RV32>)
+        {
+            XLEN instreth = getCsrRegister(INSTRETH)->dmiRead<XLEN>();
+            XLEN minstreth = getCsrRegister(MINSTRETH)->dmiRead<XLEN>();
+
+            if (SPARTA_EXPECT_FALSE(instret == std::numeric_limits<XLEN>::max()))
```
klingaard, line 655: If this is true, then won't `++instret` naturally roll over to 0?

core/PegasusState.cpp
```diff
@@ -529,6 +627,13 @@ namespace pegasus
         // for now just assume each inst takes 1 cycle
         ++sim_state_.cycles;
 
+        if (pegasus_core_->isExtensionEnabled("zicntr"))
```
klingaard: Might be faster to have a specific "already determined" boolean:
```
if (pegasus_core_->isZicntrEnabled())
```

core/PegasusState.cpp
```diff
@@ -511,7 +525,91 @@ namespace pegasus
         }
     }
 
-    template <bool CHECK_ILIMIT>
+    template <typename XLEN> void PegasusState::incrCycleCsrs_()
+    {
+        XLEN cycle = getCsrRegister(CYCLE)->dmiRead<XLEN>() + 1;
```
colby-nyce: Can you use this?
`auto cycle = PEEK_CSR_REG<XLEN>(this, CYCLE) + 1;`

Same goes for the other peeks/pokes below.

colby-nyce (approved):

core/PegasusState.cpp
kathlenemagnus: Good point! Actually I should check that instret == max BEFORE I increment it. Let me rework this.

core/PegasusState.cpp
kathlenemagnus: I could, but since these methods are called every cycle I want to use the dmi calls for performance.

core/PegasusState.cpp
knute-mips: You can always mask it too.

PR #374, opened 2026-04-14
Removed ExtensionManager from PegasusCore

kathlenemagnus:
Removed extra ExtensionManager left in PegasusCore after a recent PR, which required moving a lot of other PegasusCore methods into PegasusState. Also added new parameters to set an HART's ISA string and privilege mode when it boots.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,7 +158,7 @@ FetchContent_MakeAvailable(spike)
 ################################################################################
 
 if (PROJECT_IS_TOP_LEVEL)
-  find_program(CLANGFORMAT_EXECUTABLE NAMES clang-format-15 clang-format-16 clang-format)
+  find_program(CLANGFORMAT_EXECUTABLE NAMES clang-format)
 
   execute_process (
     COMMAND
@@ -167,7 +167,7 @@ if (PROJECT_IS_TOP_LEVEL)
        CLANGFORMAT_VERSION_OUTPUT
   )
 
-  set(REQUIRED_CLANGFORMAT_VERSION "15")
+  set(REQUIRED_CLANGFORMAT_VERSION "18")
 
   # Extract the version number from the output
   string(REGEX MATCH "[0-9]+\\.[0-9]+\\.[0-9]+" CLANGFORMAT_VERSION ${CLANGFORMAT_VERSION_OUTPUT})
@@ -188,9 +188,6 @@ if (PROJECT_IS_TOP_LEVEL)
        ${SIM_BASE}
   )
 
-  # CPUTopology is pretty clean as it is.  There's an option to NOT
-  # collapse the array of structures in clang-format v16, but we're not
-  # there yet
   add_custom_target(clangformat
     COMMAND
        ${CLANGFORMAT_EXECUTABLE} -i --files=${CMAKE_CURRENT_BINARY_DIR}/clang-format-files.out
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,7 +158,7 @@ FetchContent_MakeAvailable(spike)
 ################################################################################
 
 if (PROJECT_IS_TOP_LEVEL)
-  find_program(CLANGFORMAT_EXECUTABLE NAMES clang-format)
+  find_program(CLANGFORMAT_EXECUTABLE NAMES clang-format-15 clang-format-16 clang-format)
 
   execute_process (
     COMMAND
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,7 +158,7 @@ FetchContent_MakeAvailable(spike)
 ################################################################################
 
 if (PROJECT_IS_TOP_LEVEL)
-  find_program(CLANGFORMAT_EXECUTABLE NAMES clang-format-15 clang-format-16 clang-format)
+  find_program(CLANGFORMAT_EXECUTABLE NAMES clang-format)
 
   execute_process (
     COMMAND
--- a/core/PegasusState.cpp
+++ b/core/PegasusState.cpp
@@ -27,6 +27,22 @@
 
 namespace pegasus
 {
+    uint32_t getXlenFromIsaString(const std::string & isa_string)
+    {
+        if (isa_string.find("32") != std::string::npos)
+        {
+            return 32;
+        }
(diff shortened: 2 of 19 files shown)


kathlenemagnus:
Thank you @colby-nyce for debugging the cosim fails! That was tricky to figure out.

I updated the `PegasusState::stopSim` method to `PegasusState::setSimStopped` to be able to "unstop" simulation. In CoSim, if we flush the event that ended simulation we can now set `sim_stopped` to false and fix up the Finish Action Group to reconnect it to Fetch.

core/PegasusState.cpp
```diff
@@ -86,21 +102,42 @@ namespace pegasus
...
+        }
+
+        const std::string init_isa = p->isa;
+        if (init_isa.empty() == false)
+        {
+            isa_string_ = p->isa;
+            extension_manager_.setISA(isa_string_);
+        }
+
+        setPcAlignment_(isCompressionEnabled() ? 2 : 4);
```
colby-nyce: Should `setPcAlignment_()` take any input arguments? The value always has to be 2 or 4 and it can be determined by `isCompressionEnabled()` alone.

core/PegasusState.cpp
```diff
@@ -121,6 +158,9 @@ namespace pegasus
         add_registers(vec_rset_);
         add_registers(csr_rset_);
 
+        csr_enabled_state_.resize(csr_rset_->getNumRegisters(), true);
```
colby-nyce: Isn't this assuming that the CSR numbers are monotonically increasing from 0, like an INT/FP/VEC register? Why isn't this `std::unordered_map<uint32_t, bool> csr_enabled_states_`?

core/PegasusState.cpp
```diff
@@ -863,31 +956,44 @@ namespace pegasus
...
-        sim_state_.sim_stopped = true;
+            sim_state_.workload_exit_code = exit_code;
+            sim_state_.test_passed = (exit_code == 0) ? true : false;
 
-        finish_action_group_.setNextActionGroup(&stop_sim_action_group_);
+            finish_action_group_.setNextActionGroup(&stop_sim_action_group_);
+        }
+        else
+        {
+            (void)exit_code;
```
colby-nyce: Can you instead `sparta_assert(exit_code == 0)`?

cosim/PegasusCoSim.cpp
```diff
@@ -56,6 +56,7 @@ namespace pegasus::cosim
 
     PegasusCoSim::PegasusCoSim(uint64_t ilimit, const std::string & workload,
                                const std::map<std::string, std::string> pegasus_params,
+                               const std::vector<std::vector<std::string>> pegasus_loggers,
```
colby-nyce: `pegasus_params` and `pegasus_loggers` should be const ref

test/cosim/cosim_workload/flush_workload/FlushWorkload_test.cpp
```diff
@@ -441,7 +441,7 @@ int main(int argc, char** argv)
 
     const size_t snapshot_threshold = 10;
 
-    auto initSimConfig = [&]() -> sparta::app::SimulationConfiguration*
+    auto initSimConfig = [&](const std::string log_suffix) -> sparta::app::SimulationConfiguration*
```
colby-nyce: const ref

test/cosim/cosim_workload/flush_workload/FlushWorkload_test.cpp
```diff
@@ -451,7 +451,7 @@ int main(int argc, char** argv)
             config_truth->processParameter(param_name, param_value, false);
         }
 
-        config_truth->enableLogging("top", "inst", workload_fname + ".log");
+        config_truth->enableLogging("top", "inst", workload_fname + "." + log_suffix + ".log");
         file_cleanup.cleanupOnSuccess(workload_fname + ".log");
```
colby-nyce: `file_cleanup.cleanupOnSuccess(workload_fname + "." + log_suffix + ".log");`

test/cosim/cosim_workload/flush_workload/FlushWorkload_test.cpp
```diff
@@ -466,7 +466,7 @@ int main(int argc, char** argv)
     sparta::Scheduler scheduler_truth;
     PegasusSim cosim_truth(&scheduler_truth);
 
-    cosim_truth.configure(0, nullptr, initSimConfig());
+    cosim_truth.configure(0, nullptr, initSimConfig("truth"));
```
colby-nyce: Can you call the log suffixes "cosim_truth" and "replayer_truth"?

CMakeLists.txt
```diff
@@ -167,7 +167,7 @@ if (PROJECT_IS_TOP_LEVEL)
        CLANGFORMAT_VERSION_OUTPUT
   )
 
-  set(REQUIRED_CLANGFORMAT_VERSION "15")
+  set(REQUIRED_CLANGFORMAT_VERSION "18")
```
colby-nyce, line 170: Is this going to be added to MIPS Conda? If so, while we are changing it anyway, can we go to clang-format-19? That's what I chose as the standard for SimDB, so it'll be one less thing to install (or SimDB goes to 18, but I don't see the point).

colby-nyce:
The only concern here is the choice of `std::vector` for `csr_enabled_state_`, but other than that I'll approve. @kathlenemagnus can you comment?

core/PegasusState.cpp
kathlenemagnus: Good point. I'll remove the argument and just check if compression is enabled in the method.

CMakeLists.txt
kathlenemagnus: MIPS Conda has versions 19 and 20. I kept it at 18 because that's the highest version supported by our Ubuntu runner for the GitHub regression. I'll see if there is an updated Ubuntu version we can use, but I'm pretty sure we're using the latest release.

kathlenemagnus:
> The only concern here is the choice of `std::vector` for `csr_enabled_state_`, but other than that I'll approve. @kathlenemagnus can you comment?

I changed it from a vector to a map. Thanks for catching that! I applied the rest of your PR feedback, too.

colby-nyce (approved):

Issue #372, opened 2026-04-10
Add Support for "Zvfh" and "Zvfhmin" Extensions for Vector Half-Precision Floating-Point


kathlenemagnus:
See Chapters 31.18.4 and 31.18.5 of the RISC-V unprivileged spec. These instructions are not supported in Mavis yet, so that work will need to be completed first.

kathlenemagnus:
There are actually no new instructions that need to be added. See #376

PR #376, opened 2026-04-16
Added Support for Zvfh and Zvfhmin

kathlenemagnus:
The Zvfh/Zvfhmin extensions enable BF16 for the vector floating point instructions. There are no "new" instructions; Pegasus just needs to check whether the required extension is enabled before executing any vector floating point instructions when SEW=16.

Also added parameters to set the initial vector config, which is needed for testing single vector instructions.


kathlenemagnus:
Closes #372

klingaard (approved):

PR #384, opened 2026-05-08
adding documentation for vector extension

lzhu-pub:
All vector extension source code has been documented with Google style docstring.
It should be straightforward now to understand and use/change the existing code.
Let me know if more documentation is needed elsewhere.


core/inst_handlers/v/RvvFixedPointInsts.cpp
```diff
@@ -327,19 +327,33 @@ namespace pegasus
     template void RvvFixedPointInsts::getInstHandlers<RV32>(std::map<std::string, Action> &);
     template void RvvFixedPointInsts::getInstHandlers<RV64>(std::map<std::string, Action> &);
 
+    /**
+     * @brief Helper function to perform binary operations between vector and scalar registers.
+     *
+     * This function executes a binary operation between elements of a vector register
+     * and a scalar value from an integer register or elements of vector register. The result
```
kathlenemagnus-mips, line 334: Can you put the vector instruction suffixes in parentheses? Like `(*.vx)` and `(*.vv)`

kathlenemagnus-mips: Also put a couple of example instructions since this handler is used for a lot of instructions.

kathlenemagnus-mips: Is this method any different than `viBinaryHelper` in `RvvIntegerInsts`?

core/inst_handlers/v/RvvPermuteInsts.cpp
```diff
@@ -333,19 +360,37 @@ namespace pegasus
...
+    /**
+     * @brief Handles the RISC-V `vslide1` instruction, which performs a single-element slide
+     * operation.
+     *
+     * This function implements the `vslide1` instruction, which shifts elements within a vector
+     * register up or down by one position. The direction of the slide (up or down) is determined by
+     * the `isUp` template parameter. The operation is applied element-wise across the vector
+     * register, and the vacated position is filled with the value from the neighboring element.
+     *
+     * @tparam XLEN The width of the scalar registers (e.g., 32 or 64 bits).
```
kathlenemagnus-mips, line 393: Can you add the template param documentation for all of these methods? This is very useful.

core/inst_handlers/v/RvvFixedPointInsts.cpp
lzhu-pub: The difference is sign/zero extension. For integer instructions, spec requires number to be extended, but for fixed point, it cannot be done.
It seems that some functions implementations have been moved to header files. I will add doc for those.

kathlenemagnus-mips:
You can go ahead and merge this PR.

I'd like to also see some documentation for these files:
https://github.com/sparcians/pegasus/blob/main/core/VecElements.hpp
https://github.com/sparcians/pegasus/blob/main/core/VectorConfig.hpp

For VecElements - can you write either pseudo code or unit tests that show some basic operations? For example, how to iterate across a single register, across multiple registers when LMUL > 1, when VSTART != 0, when masking is enabled, etc.

kathlenemagnus (approved):

PR #389, opened 2026-06-22
Adds support for trace generation based on an opcode trigger

knute-mips:
Also...

1. Fixed a bug in riscv access test -- now passes.
2. Fixed a trace bug where taken branches to next PC didn't put in a taken record
3. Enhanced the symbol lookup structures.

### DIFF
--- a/core/observers/STFLogger.cpp
+++ b/core/observers/STFLogger.cpp
@@ -4,12 +4,21 @@
 #include "core/VecElements.hpp"
 #include "include/gen/pegasus_version.hpp"
 #include "core/PegasusCore.hpp"
+#include "system/PegasusSystem.hpp"
 
 namespace pegasus
 {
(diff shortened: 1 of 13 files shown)


core/observers/STFLogger.cpp
```diff
@@ -354,18 +377,25 @@ namespace pegasus
...
                                                     static_cast<uint32_t>(state->getPrivMode()));
                     stf_writer_ << stf::EventPCTargetRecord(state->getNextPc());
                 }
-                else
+                // Do not insert a PC target if the branch is not taken
+                else if (state->isBranchTaken())
                 {
+                    // Taken branch
                     stf_writer_ << stf::InstPCTargetRecord(state->getNextPc());
                 }
```
knute-mips, line 393: @bdutro is this correct?

bdutro: Yes, perfect

klingaard (approved):
Bypassing some rules here and approving my own PR.

PR #402, opened 2026-09-21
Ygunjal/fcsr and stf logger bug

yashomips:
There are 3 bugs being fixed in this PR:
1. `STFLogger` was not writing the new CSR values.
2. FCSR bits weren't implemented to be sticky, every FCSR write was clearing the previously set bits.
3. `softfloat_exceptionFlags` wasn't being reset before every op. I added the function `resetFpExceptionFlags` to `getRM` function for those which use Rounding Mode and manually clear off this variable for others which don't use `getRM`.


core/inst_handlers/f/RvfInstsBase.hpp
```diff
@@ -21,8 +21,14 @@ namespace pegasus
         using base_type = RvfInstsBase;
 
       protected:
+        static void resetFpExceptionFlags()
+        {
+            softfloat_exceptionFlags = 0;
+        }
+
```
klingaard: If we ever go multithreaded, this will need to be guarded.  But for now, this is ok.

klingaard (approved):

